PR #12, opened 2013-11-06
Making it easy to initialize the EF6 MiniProfiler

stebet:
The EF6 Miniprofiler can now be easily registered with the following
code:

// Add the using
using StackExchange.Profiling.EntityFramework6;

// Initialize EF6 MiniProfiler
MiniProfilerEF6.Initialize();

### DIFF
--- a/StackExchange.Profiling.EntityFramework6/MiniProfilerEF6.cs
+++ b/StackExchange.Profiling.EntityFramework6/MiniProfilerEF6.cs
@@ -1,0 +1,65 @@
+﻿namespace StackExchange.Profiling.EntityFramework6
+{
+    using System;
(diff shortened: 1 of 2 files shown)


ctolkien:
Hitting the exact same issue. I've removed and re-installed, double checked the binding redirect, etc. Still not working.

```
<dependentAssembly>
        <assemblyIdentity name="MiniProfiler" publicKeyToken="b44f9351044011a3" culture="neutral" />
        <bindingRedirect oldVersion="0.0.0.0-3.2.0.157" newVersion="3.2.0.157" />
      </dependentAssembly>
```

And calling `MiniProfilerEF6.Initialize();` as the very first line of Application_Start

PR #22, opened 2013-12-04
Adding compression to results

SeriousM:
The output of the MiniProfilerHandler is now compressed, if the request supports that.
Added unit tests to test the behaviour of the compression.

This would significantly reduce the transmitted information because:
- "results" is sending lists of guids all the time
- "includes.css", "includes.js" and "includes.tmpl" is reduced in size

I took care about clients that does not support the compression at all.

### DIFF
--- a/StackExchange.Profiling.Tests/MiniProfileHandlerTests.cs
+++ b/StackExchange.Profiling.Tests/MiniProfileHandlerTests.cs
@@ -1,4 +1,6 @@
 ﻿using System;
+using System.Collections.Specialized;
+using System.IO.Compression;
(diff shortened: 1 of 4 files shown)


thdk:
We are compressing all scripts ourselves. Would be great to have a setting to disable compression.

yellis:
@thdk Want to submit a PR? Add a static property to MiniProfiler.Settings.cs to control it.

thdk:
@yellis I'm doing this now. Although new to git.. I'll give it a shot :)

thdk:
PR submitted.

PR #122, opened 2016-01-07
Mongocsharpdriver2.2

claude-uceda:
First time I contribute, sorry if I did incorrectly


NickCraver:
Since this has breaking changes, it'd have to be reserved for a major version release. Most of the PR looks sound, but actually merging may be a pain. I think we're likely looking at forking the library for the .Net MVC Core world here shortly, and maintaining 2 branches. I'll see what we can do about pulling in the Mono update, but requiring users to jump from .Net 4.0 is another breaking gap we'd have to do in a major version.

NickCraver:
Going to close this out in favor of #125, which has *far* more changes involved for v4. Thanks for the effort here, and I do apologize this has sat for so long :(

claude-uceda:
No problem, thanks for letting me know.  

Cheers

PR #124, opened 2016-01-18
[Closes #107] Add RavenDB 3.0 compatibility

### LINKED ISSUE #107: Update RavenDB to 3.0
Need to update RavenDB support for 3.0. Maybe just as separate project (MiniProfiler.RavenDb30) since it was a major upgrade.

kamranayub:
For issue #107.

I updated the `RavenMiniProfiler` implementation to be compatible with the new Async server client in Raven 3.0. This also maintains backwards compatibility with RavenDB 2.5 since the API is the same.

In Raven 2.5, all requests were synchronous so `HttpContext.Current` was accessible from the `LogRequest` event. However, in 3.0, everything is async-first, so more often than not, when `LogRequest` is thrown, it's on a different thread context which does not have access to `HttpContext.Current` so it could never find `MiniProfiler.Current`.

It's a little awkward but essentially this handles a different `ConfigureRequest` event which is thrown on every Raven request and is *before* any request is made so `HttpContext.Current` is available. We create a handler we pass that context/profiler into the `IncludeTiming` function to add a timing. Once the timing is added, we unsubscribe from the request since we aren't sure we can reuse the same context.

I also discovered a couple bugs that I addressed:

- Sometimes accessing MiniProfiler.Current can result in a NRE even when checking null (I'm not 100% sure what caused this)
- Not every `ConfigureRequest` event has a corresponding `LogRequest` event. I was seeing, on app initialization, about 14 duplicate timings for the same request. To address this I just added a simple marker using a property bag RavenDB associates with each request passed around.
- **Not Solved**: Simultaneous requests might accidentally write to separate HttpContexts, I believe due to how the event is being fired. Still investigating a fix.

### DIFF
--- a/StackExchange.Profiling.RavenDb/RavenMiniProfiler.cs
+++ b/StackExchange.Profiling.RavenDb/RavenMiniProfiler.cs
@@ -8,7 +8,9 @@ namespace StackExchange.Profiling.RavenDb
 
     public class MiniProfilerRaven
     {
-        private static readonly Regex IndexQueryPattern = new Regex(@"/indexes/[A-Za-z/]+");
+        private const string RavenHandledRequestMarker = "__MiniProfiler.Raven_handled";
+
+        private static readonly Regex IndexQueryPattern = new Regex(@"/indexes/[A-Za-z/]+");        
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


NickCraver:
Thanks for the fixes!

PR #129, opened 2016-05-19
[Feature] Use v4.5 and async sql profiling

jonrad:
Add 4.5 Configurations
Make the sql profiler work with 4.5 async functions


NickCraver:
I think you caught us at a really bad time here because this library is coming up after a few others for a very drastic change for .NET Core and MVC 1 support. The project file changes here are the reverse of what happens in the new project system, though the actual code bits look good on first pass.

I don't want to merge *this* pull request as-is due to the project complications, but I will try and pull in the commit for the code pieces when making a `net45` (and other) builds in the new `.nupkg` layout.

NickCraver:
I'm working in the v4 branch now, if you wouldn't mind doing this changes on top of it (the 2 files are just tests and the `ProfiledDbCommand`, projects need no love now), I'd really appreciate it. You've done the work here and I've love for the git commit history to give proper credit. Any chance you'd have time to apply these changes (additions) to that branch in a PR?

NickCraver:
@jonrad  I want to pull this into the v4 branch, but I'm afraid I'll lose your name on the commits with the amount of rebasing I'll have to do. Is there any chance you can add the Async profiling overloads/tests in there? If not, I'll try my best to move them over, but the amount of moving is likely going to be problematic.

Note: most/all of the #if checks aren't needed - since MP v4 will support .NET 4.6 and up, greatly simplifying things. The new solution is on VS 2017 (it's testing the new `.csproj` system), please let me know if that's a problem.

NickCraver:
There was actually quite a bit more here - did a comprehensive commit covering the async bases here: 8973afb7bd418a37ce38e0805dc3636942ea7788

Closing out to cleanup, async should be much, much nicer to look at now.

PR #133, opened 2016-06-16
Make use of AuthorizeRequest to ensure a 401 status code for single profiler result

TheCloudlessSky:
If a non-authorized user visits a MiniProfiler results url (e.g. `http://app.com/mini-profiler-resources/results?id=<guid>`), they do *not* get a 401 status code. Having a 401 status code is what the documentation for `MiniProfiler.Settings.Results_Authorize` says it should return.


yellis:
looks good, but you are missing a closing parentheses on the if line (please squash to one commit)

TheCloudlessSky:
@yellis Fixed and squashed.

PR #135, opened 2016-06-17
Support async, fix timing level error

pengweiqhca:
Support async now

xxxx at 2016/6/17 6:57:49
 Locking = 95ms
> outer = 89.7ms
>> step 3 = 32.6ms
>>> sub-step3 = 13ms
>> step 0 = 32.4ms
>>> sub-step0 = 13.2ms
>> step 2 = 32.5ms
>>> sub-step2 = 13ms
>> step 1 = 32.5ms
>>> sub-step1 = 12.9ms
>> step 4 = 32.3ms
>>> sub-step4 = 13ms


NickCraver:
Using CallContext assumes context is preserved across the async boundary, so the fundamental approach here, unfortunately, isn't sound.

I'm finally starting to take a look at this today though, we'll see what we can come up with for options. Since there are other *major* changes to make this work, we should likely close out all PRs on it - but I want to (finally) give people a heads up.

Issue #127, opened 2016-04-28
Support for async/await


seesharper:
Hi!

I'm considering using the MiniProfiler in a self hosted Web API application that naturally uses a lot of async methods. I'm just wondering if I could do something like this?

```c#
public class AsyncLocalProfilerProvider : BaseProfilerProvider
{
    readonly AsyncLocal<MiniProfiler> currentProfiler = new AsyncLocal<MiniProfiler>();

    [Obsolete("Please use the Start(string sessionName) overload instead of this one. ProfileLevel is going away.")]
    public override MiniProfiler Start(ProfileLevel level, string sessionName = null)
    {
        return Start(sessionName);
    }

    public override MiniProfiler Start(string sessionName = null)
    {
        var profiler = new MiniProfiler(sessionName);
        SetProfilerActive(profiler);
        currentProfiler.Value = profiler;
        return profiler;
    }

    public override void Stop(bool discardResults)
    {
        var profiler = currentProfiler.Value;
        if (profiler == null)
        {
            StopProfiler(profiler);
            currentProfiler.Value = null;
        }
    }

    public override MiniProfiler GetCurrentProfiler()
    {
        return currentProfiler.Value;
    }       
}
```

NickCraver:
It's a little more involved here, since MiniProfiler doesn't need to be AsyncLocal so much as the inner `Head` does. But, the v4 branch now takes care of this (and does so by default). You'll just be able to use MiniProfiler and have async handled correctly with the upcoming release.

Closing this out just to clean up, but happy to answer any questions!

PR #140, opened 2016-09-27
Add MiniProfilerConcurrencyTest for testing concurrent profiling with threads and tasks

anjdreas:
Inspired by #127.
This is my attempt to get MiniProfiler working with async/await and generally multi-threading.
- Add `CallContextProfilerProvider` to store `Current` and `Head` separately for async/await call contexts
- Add lock() around `StepImpl()` for thread-safety, was not an issue with Task-only code
- Add tests for both simulated time and real time using tasks and threads
- Client code must assign instance to `MiniProfiler.Settings.ProfilerProvider` to override the default

I would love to get some feedback on this. It seems to work now, but not tested thoroughly yet.

Example usage:

``` csharp
[Test]
public async Task Test()
{
    MiniProfiler.Settings.ProfilerProvider = new CallContextProfilerProvider();
    var profiler = MiniProfiler.Start("root");

    // Add 100ms to root
    await Task.Delay(100);

    // 100ms + 100ms = 200ms
    var step1 = Task.Run(async () =>
    {
        using (profiler.Step("step1.0"))
        {
            await Task.Delay(100);

            await Task.Run(async () =>
            {
                using (profiler.Step("step1.1"))
                {
                    await Task.Delay(100);
                }
            });
        }
    });

    // 100ms
    var step2 = Task.Run(async () =>
    {
        using (profiler.Step("step2.0"))
        {
            await Task.Delay(100);
        }
    });

    // Longest task is 200ms
    await Task.WhenAll(step1, step2);

    MiniProfiler.Stop();

//  ANDREAST460P at 2016-09-27 12:19:08
//   root = 330.9ms
//  > step2.0 = 107.6ms
//  > step1.0 = 212.1ms
//  >> step1.1 = 107ms
    Console.WriteLine(profiler.RenderPlainText());
}
```


anjdreas:
I just realized a shortcoming, calling `MiniProfiler.Start()` multiple times down an async method chain will erase the previous one, instead of connecting to it as a child session. I think the impl in #127 had some code to handle this that I did not understand at the time. Will revisit that.

NickCraver:
I think what's likely to happen here for an async world is what we were waiting on: `AsyncLocal<T>` for context passing. We're going to look at this for .NET Core today and take a peek. I'm unsure how much will get done, but the actual implementation here is, unfortunately, on top of .NET Code bugs from some framework versions. In .NET 4.6+ it's resolved though, so I'm tempted to just say "it works on .NET 4.6 and up" from a pure sanity standpoint, especially since we'll have to support 2 libraries long-term.

anjdreas:
Thanks for looking into the topic.
In the meantime I have been successfully using https://github.com/ef-labs/nanoprofiler to do async profiling and generating reports.

NickCraver:
Hey @anjdreas,
I've done the initial async commits and some handling to MiniProfiler - the approach is different and the project system is very much different (VS 2017) though. Is there any chance you can commit those tests in a PR against the v4 branch? I certainly want your name to be on them, and if I squash across the crazy, that won't happen which is uncool.

If not I'll try and make it work, but if you had time to add them I'd very much appreciate it! If they don't work - I'll get to fixing :)

anjdreas:
Hi @NickCraver, that is awesome.

I rebased the branch on top of `v4` and removed the custom `CallContextProfilerProvider` stuff to use the built-in profiler instead.

All three tests in `MiniProfilerConcurrencyTest` now pass, but the `_RealTime` tests are a bit flaky and I'm not really sure how they can be rewritten to be more reliable. To my knowledge, threads and tasks are difficult if not impossible to do deterministic real-time profiling on.

Let me know if these tests provide any value to you and if you need anything else.

anjdreas:
I just noticed this PR is still against master, will create a new PR if you want to proceed with this.

NickCraver:
@anjdreas absolutely - I'd love to have these tests in, can you please do it on the V4 branch? I *think* all that needs adding is new test files to the project (new clean PR is easier) - happy to help just want your name on your work!

anjdreas:
Closed in favor of #149

Issue #35, opened 2014-02-17
Timing.Children is not threadsafe. 


yellis:
See [related thread on the community site](http://community.miniprofiler.com/t/timing-class-is-not-threadsafe/30?u=yaakov_ellis).

Want to make interactions with `Timing.Children` threadsafe.

alexsku:
I think I've run into the Children concurrency issue using MiniProfiler V4. This is a PR illustrating the issue (based on V4 MVC sample): https://github.com/alexsku/dotnet/pull/4

In my case I have multiple async tasks executing and awaited by the calling thread all together (using Task.WhenAll). Each task is using await with ConfigureAwait(false) (a typical pattern for a service layer in a web application). Because of the ConfigureAwait(false) the continuations for the tasks will be executed in parallel causing the concurrency issue.

I think using lock is a good way to solve the concurrency issue. Here is my PR that is addressing it (among other things): https://github.com/MiniProfiler/dotnet/pull/147 (the PR is for V4 of MiniProfiler). 

I think lock's performance is pretty good when there is low contention (which i think is the case with MiniProfiler usages, otherwise this issue would be very high priority). Here is a small test where lock is measured against other synchronization primitives: https://github.com/alexsku/dotnet/pull/5 when the lock is not contended. 

Btw, I think V4 fixes the issue described by @dixon above by using AsyncLocal

NickCraver:
Fixed by 9ec58fa5a323c7c4b10a07deb615afdcfaed8abd

PR #147, opened 2016-12-22
async multithreading support

alexsku:
This is my attempt to improve multithreading support for miniprofiler v4. I think the current version of the mini profiler would have issues in scenarios where the current thread spawns more than one task and waits for them.
```
        [Fact]
        public void Head_WhenMultipleTasksSpawned_EachSetsItsOwnHead()
        {
            MiniProfiler.Settings.ProfilerProvider = new DefaultProfilerProvider();
            var allTasks = new SemaphoreSlim(0, 1);
            var completed = new TaskCompletionSource<int>();
            using (GetRequest("http://localhost/Test.aspx", startAndStopProfiler: false))
            {
                MiniProfiler.Start();
                var sut = MiniProfiler.Current;
                var head = sut.Head;

                Task.Run(() => {
                    Assert.Equal(head, sut.Head);
                    using (sut.Step("test1"))
                    {
                        allTasks.Release();
                        completed.Task.Wait();
                    }
                });
                allTasks.Wait();
                Task.Run(() => {
                    using (sut.Step("test2"))
                    {
                        allTasks.Release();
                        completed.Task.Wait();
                    }
                });
                allTasks.Wait();

                Assert.NotNull(sut.Head);
                Assert.Equal(head, sut.Head);
                completed.SetResult(0);
            }
        }
```
In this scenario there would be multiple threads working with the mini profiler (since they would share it via AsyncLocal). This could lead to random failures since Timing class is not thread safe. Also this could lead to inconsistencies since the spawned tasks would replace the Head timing for the mini profiler instance, with one of the tasks's timing potentially becoming a child of another's.

### DIFF
--- a/src/MiniProfiler.Providers.SqlServer/SqlServerStorage.cs
+++ b/src/MiniProfiler.Providers.SqlServer/SqlServerStorage.cs
@@ -85,15 +85,17 @@ where not exists (select 1 from MiniProfilers where Id = @Id)"; // this syntax w
                 }
 
                 SaveTimings(timings, conn);
-                if (profiler.ClientTimings != null && profiler.ClientTimings.Timings != null && profiler.ClientTimings.Timings.Any())
+
+                var clientTimings = profiler.ClientTimings?.Timings;
+                if (clientTimings != null && clientTimings.Any())
                 {
                     // set the profilerId (isn't needed unless we are storing it)
-                    profiler.ClientTimings.Timings.ForEach(x =>
+                    foreach(var x in clientTimings)
                     {
                         x.MiniProfilerId = profiler.Id;
                         x.Id = Guid.NewGuid();
-                    });
-                    SaveClientTimings(profiler.ClientTimings.Timings, conn);
+                    }
+                    SaveClientTimings(clientTimings, conn);
                 }
             }
         }
@@ -139,7 +141,7 @@ WHERE NOT EXISTS (SELECT 1 FROM MiniProfilerTimings WHERE Id = @Id)";
             }
         }
 
-        private void SaveClientTimings(List<ClientTiming> timings, DbConnection conn)
+        private void SaveClientTimings(IReadOnlyList<ClientTiming> timings, DbConnection conn)
         {
             const string sql = @"
 INSERT INTO MiniProfilerClientTimings
--- a/src/MiniProfiler.Shared/ClientTimings.cs
+++ b/src/MiniProfiler.Shared/ClientTimings.cs
@@ -14,12 +14,26 @@ namespace StackExchange.Profiling
     {
         private const string TimingPrefix = "clientPerformance[timing][";
         private const string ProbesPrefix = "clientProbes[";
+        private readonly object _lockObject = new object();
+        private List<ClientTiming> _timings;
 
         /// <summary>
         /// Gets or sets the list of client side timings
         /// </summary>
         [DataMember(Order = 2)]
-        public List<ClientTiming> Timings { get; set; }
+        public List<ClientTiming> Timings
+        {
+            get
+            {
+                lock(_lockObject)
+                    return _timings == null ? null : new List<ClientTiming>(_timings);
+            }
+            set
+            {
+                lock (_lockObject)
+                    _timings = value == null ? null : new List<ClientTiming>(value);
+            }
+        }
 
         /// <summary>
         /// Gets or sets the redirect count.
--- a/src/MiniProfiler.Shared/Helpers/Net45/AsyncLocal.cs
+++ b/src/MiniProfiler.Shared/Helpers/Net45/AsyncLocal.cs
@@ -1,0 +1,78 @@
+﻿#if NET45
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Runtime.Remoting.Messaging;
+using System.Runtime.Serialization;
+
+namespace StackExchange.Profiling.Helpers.Net45
+{
+    /// <summary>
+    /// Implements interface similar to AsyncLocal which is not available until .Net 4.6.
+    /// 
+    /// The implementation is inspired by Ambient Context Logic from here: 
+    /// https://github.com/mehdime/DbContextScope/blob/master/Mehdime.Entity/Implementations/DbContextScope.cs
+    /// Unlike DbContextScope's implementation this implementation doesn't support serialization/deserialization 
+    /// ths not allowing to cross app domain barrier.
+    /// </summary>
+    internal class AsyncLocal<T>
+    where T : class
+    {
+        public AsyncLocal()
+        {
+        }
+
+        public AsyncLocal(
+            T obj
+            )
+        {
+            Value = obj;
+        }
+
+        /// <summary>
+        /// Gets or Sets the value.
+        /// </summary>
+        public T Value
+        {
+            get
+            {
+                return (CallContext.LogicalGetData(_id) as ObjectRef)?.Ref;
+            }
+
+            set
+            {
+                CallContext.LogicalSetData(_id, new ObjectRef { Ref = value });
+            }
+        }
+
+        /// <summary>
+        /// Identifies this instance of <see cref="AsyncLocal"/> in CallContext.
+        /// </summary>
+        private readonly string _id = Guid.NewGuid().ToString();
+
+        [Serializable]
+        private class ObjectRef : MarshalByRefObject, ISerializable
+        {
+            // The special constructor is used to deserialize values.
+            public ObjectRef()
+            {
+            }
+
+            // The special constructor is used to deserialize values.
+            public ObjectRef(
+                SerializationInfo info,
+                StreamingContext context
+                )
+            {
+            }
+
+            public void GetObjectData(SerializationInfo info, StreamingContext context)
+            {
+            }
+
+            public T Ref { get; set; }
+        }
+    }
+}
+#endif
--- a/src/MiniProfiler.Shared/MiniProfiler.Shared.csproj
+++ b/src/MiniProfiler.Shared/MiniProfiler.Shared.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
+﻿<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
   <PropertyGroup>
     <TargetFrameworks>net45;netstandard1.5</TargetFrameworks>
     <AssemblyName>MiniProfiler.Shared</AssemblyName>
--- a/src/MiniProfiler.Shared/MiniProfiler.Shared.csproj
+++ b/src/MiniProfiler.Shared/MiniProfiler.Shared.csproj
@@ -22,6 +22,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="NETStandard.Library" Version="1.6.1" />
+    <PackageReference Include="System.Collections.Immutable" Version="1.3.0" />
   </ItemGroup>
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard1.5'">
     <PackageReference Include="System.ComponentModel.Primitives" Version="4.3.0" />
--- a/src/MiniProfiler.Shared/MiniProfiler.cs
+++ b/src/MiniProfiler.Shared/MiniProfiler.cs
@@ -135,10 +135,9 @@ namespace StackExchange.Profiling
                     {
                         var timing = timings.Pop();
 
-                        if (timing.HasChildren)
+                        var children = timing.Children;
+                        if (children?.Count > 0)
                         {
-                            var children = timing.Children;
-
                             for (int i = children.Count - 1; i >= 0; i--)
                             {
                                 children[i].ParentTiming = timing;
@@ -189,7 +188,19 @@ namespace StackExchange.Profiling
         /// <summary>
         /// Gets or sets points to the currently executing Timing. 
         /// </summary>
-        public Timing Head { get; set; }
+        public Timing Head
+        {
+            get
+            {
+                Settings.EnsureProfilerProvider();
+                return Settings.ProfilerProvider.CurrentHead;
+            }
+            set
+            {
+                Settings.EnsureProfilerProvider();
+                Settings.ProfilerProvider.CurrentHead = value;
+            }
+        }
 
         /// <summary>
         /// Gets the ticks since this MiniProfiler was started.
@@ -329,9 +340,9 @@ namespace StackExchange.Profiling
 
                 yield return timing;
 
-                if (timing.HasChildren)
+                var children = timing.Children;
+                if (children?.Count > 0)
                 {
-                    var children = timing.Children;
                     for (int i = children.Count - 1; i >= 0; i--) timings.Push(children[i]);
                 }
             }
--- a/src/MiniProfiler.Shared/ProfileProviders/BaseProfilerProvider.cs
+++ b/src/MiniProfiler.Shared/ProfileProviders/BaseProfilerProvider.cs
@@ -29,6 +29,11 @@ namespace StackExchange.Profiling
         public abstract MiniProfiler GetCurrentProfiler();
 
         /// <summary>
+        /// GEts or sets the current head timing. This is used by <see cref="MiniProfiler.Head"/>.
+        /// </summary>
+        public abstract Timing CurrentHead { get; set; }
+
+        /// <summary>
         /// Sets <paramref name="profiler"/> to be active (read to start profiling)
         /// This should be called once a new MiniProfiler has been created.
         /// </summary>
--- a/src/MiniProfiler.Shared/ProfileProviders/DefaultProfilerProvider.cs
+++ b/src/MiniProfiler.Shared/ProfileProviders/DefaultProfilerProvider.cs
@@ -1,7 +1,8 @@
 ﻿#if NET45
 using System;
 using System.Runtime.Remoting.Messaging;
-using System.Web;
+using System.Web; 
+using StackExchange.Profiling.Helpers.Net45;
 #else
 using System.Threading;
 #endif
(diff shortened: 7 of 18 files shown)


src/MiniProfiler.Shared/Helpers/Net45/AsyncLocal.cs
```diff
@@ -0,0 +1,78 @@
...
+{
+    /// <summary>
+    /// Implements interface similar to AsyncLocal which is not available until .Net 4.6.
+    /// 
+    /// The implementation is inspired by Ambient Context Logic from here: 
+    /// https://github.com/mehdime/DbContextScope/blob/master/Mehdime.Entity/Implementations/DbContextScope.cs
+    /// Unlike DbContextScope's implementation this implementation doesn't support serialization/deserialization 
+    /// ths not allowing to cross app domain barrier.
+    /// </summary>
+    internal class AsyncLocal<T>
```
alexsku: a polyfill for AsyncLocal which is not available before .Net 4.6

src/MiniProfiler.Shared/MiniProfiler.cs
```diff
@@ -189,7 +188,19 @@ public Timing Root
         /// <summary>
         /// Gets or sets points to the currently executing Timing. 
         /// </summary>
-        public Timing Head { get; set; }
+        public Timing Head
+        {
+            get
+            {
+                Settings.EnsureProfilerProvider();
+                return Settings.ProfilerProvider.CurrentHead;
```
alexsku: It is possible that the current thread spawns more than one task on separate threads and wait for them. As a result technically it is possible that there would be more than one separate head for a miniprofiler instance (one for each spawned thread). This code offloads managing heads to the profiler provider, so if the provider handles spawning threads (i.e. it relies on AsyncLocal or CallContext) then it should be able to manage head properly.

src/MiniProfiler.Shared/ProfileProviders/DefaultProfilerProvider.cs
```diff
@@ -1,7 +1,8 @@
 ﻿#if NET45
 using System;
 using System.Runtime.Remoting.Messaging;
-using System.Web;
+using System.Web; 
+using StackExchange.Profiling.Helpers.Net45;
```
alexsku: Since there is AsyncLocal polyfill we can keep the implementation of the provider the same - it will be either using the polyfill or .net version of AsyncLocal

src/MiniProfiler.Shared/Timing.cs
```diff
@@ -77,28 +78,76 @@ public Timing(MiniProfiler profiler, Timing parent, string name, decimal? minSav
         /// Gets or sets All sub-steps that occur within this Timing step. Add new children through <see cref="AddChild"/>
         /// </summary>
         [DataMember(Order = 5)]
-        public List<Timing> Children { get; set; }
+        public List<Timing> Children
```
alexsku: since technically there could be multiple threads working with the same Timing instance at the same time (my example above with a thread spawning two tasks and waiting for them) Children and CustomTimings need to be protected from multithreaded access. Here I added a backing field _children and protected it with _lockObject. The property always return a copy of _children so the caller can iterate through the list safely. I didn't see any code that would attempt to modify the children of a Timing via Children property so I think nothing is broken by this change

alexsku: Ideally the property should return IReadOnlyList<Timing> (by just casting the created List) but protobuf doesn't really work well with IReadOnlyList<Timing> (due to this issue: https://github.com/mgravell/protobuf-net/issues/90). It might be possible to use IImmutableList<Timing> (backed by ImmutableArray) which is supported by protobuf. But the problem would be with CustomTimings property which returns Dictionary. IReadOnlyDictionary would not work with protobuf either, and ImmutableDictionary has significantly worse performance comparing to Dictionary which makes me reluctant to use it. 

Another option is to change the property to GetChildren/SetChildren method pair to emphasize that there are not just access to the backing filed but actually copy the contents of it.

Please let me know what you prefer I can change it

src/MiniProfiler.Shared/Timing.cs
```diff
@@ -77,28 +78,76 @@ public Timing(MiniProfiler profiler, Timing parent, string name, decimal? minSav
...
+                }
+            }
+        }
 
         /// <summary>
         /// <see cref="CustomTiming"/> lists keyed by their type, e.g. "sql", "memcache", "redis", "http".
         /// </summary>
         [DataMember(Order = 6)]
-        public Dictionary<string, List<CustomTiming>> CustomTimings { get; set; }
+        public Dictionary<string, List<CustomTiming>> CustomTimings
```
alexsku: Similarly to Children CustomTimings is backed by a field (_customTimings) and is protected with _lockObject

src/MiniProfiler/ProfileProviders/WebRequestProfilerProvider.cs
```diff
@@ -150,7 +150,23 @@ private static void EnsureName(MiniProfiler profiler, HttpRequest request)
         /// </summary>
         public override MiniProfiler GetCurrentProfiler() => Current;
 
+        /// <summary>
+        /// Current head timing.
+        /// </summary>
+        public override Timing CurrentHead
```
alexsku: I'm not sure what is the intended future of WebRequestProfilerProvider and why DefaultProfilerProvider cannot be used all the time. I added this property only to support the changes to the interface. 

Another option is to "merge"  WebRequestProfilerProvider and  DefaultProfilerProvider, in this case WebRequestProfilerProvider.Current and WebRequestProfilerProvider.CurrentHead would be backed by AsyncLocals (or the polyfills). I don't see specific downside with this and the upside is that this provider would support async controller actions (that are using .ConfigureAwait(false) internally).

tests/MiniProfiler.Tests/App.config
```diff
@@ -34,4 +34,7 @@
       <add name="SQLite Data Provider" invariant="System.Data.SQLite" description=".NET Framework Data Provider for SQLite" type="System.Data.SQLite.SQLiteFactory, System.Data.SQLite" />
     </DbProviderFactories>
   </system.data>
+  <appSettings>
+    <add key="xunit.parallelizeTestCollections" value="false"/>
```
alexsku: it looks like xunit prefers to run unit tests in parallel, which causes random issues when static fields (e.g. ProfilerProvider) are getting changed. I think the best is to disable parallelization, i think this disables it (https://xunit.github.io/docs/configuring-with-xml.html)

tests/MiniProfiler.Tests/BaseTest.cs
```diff
@@ -173,7 +173,7 @@ protected void AssertPublicPropertiesAreEqual<T>(T t1, T t2)
             // we'll handle any collections elsewhere
             var props = from p in typeof(T).GetProperties(BindingFlags.Public | BindingFlags.Instance)
                         where p.IsDefined(typeof(System.Runtime.Serialization.DataMemberAttribute), false)
-                        && !p.PropertyType.GetInterfaces().Any(i => i.Equals(typeof(IDictionary)) || i.Equals(typeof(IList)))
+                        && !p.PropertyType.GetInterfaces().Any(i => i.Equals(typeof(IEnumerable)))
```
alexsku: I think most of implementations of enumerable cannot be compared using Object.Equals, so i think this is the best catch all case for all enumerable (otherwise every time i try to use IImmutableList the tests fail since is not an IList)

tests/MiniProfiler.Tests/MiniProfilerTest.cs
```diff
@@ -33,6 +36,102 @@ public void Simple()
...
+
+                await Task.Delay(TimeSpan.FromMilliseconds(1)).ConfigureAwait(configureAwait);
+
+                Assert.NotNull(sut.Head);
+                Assert.Equal(head, sut.Head);
+            }
+        }
+
+        [Fact]
+        public void Head_WhenMultipleTasksSpawned_EachSetsItsOwnHead()
```
alexsku: this test is a demonstration of the scenario i'm trying to support when one thread spawns two tasks and wait for their completion. In this scenario I would expect the tasks's Timings to be children of the main thread's Timing instance

tests/MiniProfiler.Tests/Storage/SqlServerStorageTest.cs
```diff
@@ -16,6 +16,7 @@ public class SqlServerStorageTest : BaseTest, IClassFixture<SqlCeStorageFixture<
         
         public SqlServerStorageTest(SqlCeStorageFixture<SqlServerStorageTest> fixture)
         {
+            MiniProfiler.Settings.ProfilerProvider = new WebRequestProfilerProvider();
```
alexsku: the test should be using WebRequestProfilerProvider the DefaultProfilerProvider doesn't save the stopped miniprofiler instances and fails these tests.

src/MiniProfiler.Shared/ClientTimings.cs
```diff
@@ -14,12 +14,26 @@ public class ClientTimings
...
         /// Gets or sets the list of client side timings
         /// </summary>
         [DataMember(Order = 2)]
-        public List<ClientTiming> Timings { get; set; }
+        public List<ClientTiming> Timings
+        {
+            get
+            {
+                lock(_lockObject)
+                    return _timings == null ? null : new List<ClientTiming>(_timings);
```
NickCraver: This creates a duplicate list *on every access*, that's a lot of allocations for the safety.

src/MiniProfiler.Shared/Timing.cs
NickCraver: This approach (like earlier comments) creates *a lot* of allocations. A new list per accessor really just isn't tenable from a performance or goals standpoint for the library. I'm just not sure any of the proposed solutions is lightweight enough to work.

src/MiniProfiler.Shared/Timing.cs
NickCraver: Same as comments above - many copies and *a lot* of overhead added here for only a little bit of safety.

Also note that simply by the locking nature here, MiniProfiler would be *causing* performance problems itself. That's definitely a no-no for any profiler.

src/MiniProfiler/ProfileProviders/WebRequestProfilerProvider.cs
NickCraver: I agree it's possible WebRequest provider can go away, but it's a fairly drastic behavior change and possible weight and breaking change to do such. Consider at the moment the case where there is no profiler because of the async context breaks to backend code. If suddenly there is a profiler on massive operations, that may add significant weight and allocations (even from a stack trace strings standpoint), as well as serialization costs at the end to that path. This can easily break things.

I'm not against it per se, but it'd definitely be a top breaking change alert in the v4 notes.

tests/MiniProfiler.Tests/App.config
NickCraver: The proper fix here is to categorize the tests, which is already happening in some cases. Disabling parallelization just hides actual issues.

src/MiniProfiler.Providers.SqlServer/SqlServerStorage.cs
```diff
@@ -85,15 +85,17 @@ public override void Save(MiniProfiler profiler)
...
                     // set the profilerId (isn't needed unless we are storing it)
-                    profiler.ClientTimings.Timings.ForEach(x =>
+                    foreach(var x in clientTimings)
                     {
                         x.MiniProfilerId = profiler.Id;
                         x.Id = Guid.NewGuid();
-                    });
-                    SaveClientTimings(profiler.ClientTimings.Timings, conn);
+                    }
+                    SaveClientTimings(clientTimings, conn);
```
NickCraver: +1 - good cleanup cleanup on these.

src/MiniProfiler.Providers.SqlServer/SqlServerStorage.cs
```diff
@@ -85,15 +85,17 @@ public override void Save(MiniProfiler profiler)
                 }
 
                 SaveTimings(timings, conn);
-                if (profiler.ClientTimings != null && profiler.ClientTimings.Timings != null && profiler.ClientTimings.Timings.Any())
+
+                var clientTimings = profiler.ClientTimings?.Timings;
+                if (clientTimings != null && clientTimings.Any())
```
NickCraver: `clientTiming?.Any() ?? false`

src/MiniProfiler.Shared/Timing.cs
alexsku: Do you think there will be many accesses to the Children property? While collecting timings during the profiler session Children property is no longer used, instead the backing list is used, where the reallocations do not happen. I think the only time when Children property is accessed is during the processing of captured mini profiler snapshot. And even there i wouldn't expect many of the accesses. Maybe I'm missing something, could you give me examples that you think where the property will be accessed a lot?

src/MiniProfiler.Shared/Timing.cs
alexsku: I don't think locks add much overhead actually. Most of the time it would be uncontested locking where the overhead is fairly minimal (this is an example i found someone attempted to measure it: http://kynosarges.org/LockPerformance.html) and when the locks are contested clearly we want to have them. 

Also I don't think there is any alternative to using locks. As soon as the code starts using DefaultProfileProvider (from the v4 branch) there is a possibility of multiple threads accessing single Timing, the only thing the application needs to do is to send requests to a couple of data sources (e.g. redis and database) asynchronously and wait for them all at one place (like using Task.WhenAll). I'm not sure how you envision to support mitigate this without using locks.

src/MiniProfiler/ProfileProviders/WebRequestProfilerProvider.cs
alexsku: I think maybe the behavior change concern can be mitigated by introducing AsyncWebRequestProfilerProvider that would support the async operations while WebRequestProfilerProvider can stay unchanged. As a result the switch to the new behavior would be an explicit action required from the user of the library

tests/MiniProfiler.Tests/App.config
alexsku: this is true, however i think this might be pretty tough to do with the mini profiler library tests. The library relies on a static variable to keep the profiler provider and some of the tests replace profiler providers, if a provider is replaced in a middle of test's execution it might result in some fairly weird failures. I think pretty much all tests would need to be in the same category and in future it would be a pain to maintain the categorization, since the requirements are not quite obvious. 

And even with the parallelization disabled the unit tests execute pretty fast.

src/MiniProfiler.Shared/MiniProfiler.Shared.csproj
```diff
@@ -22,6 +22,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="NETStandard.Library" Version="1.6.1" />
+    <PackageReference Include="System.Collections.Immutable" Version="1.3.0" />
```
alexsku: Oops, i should remove it, it is a left over from my experiments with using immutable collections

src/MiniProfiler.Shared/ProfileProviders/BaseProfilerProvider.cs
```diff
@@ -29,6 +29,11 @@ public abstract class BaseProfilerProvider : IProfilerProvider
         public abstract MiniProfiler GetCurrentProfiler();
 
         /// <summary>
+        /// GEts or sets the current head timing. This is used by <see cref="MiniProfiler.Head"/>.
+        /// </summary>
+        public abstract Timing CurrentHead { get; set; }
```
alexsku: I understand your concern, my thinking was that there could be more than one head timing in an async scenario. However in the current design whether async scenario is supported is left up to the profiler provider, so it looks like the only way to support is to store current head in the profiler provider. Do you think it would be better to move the AsyncLocal backing the head into mini profiler?

src/MiniProfiler.Shared/Timing.cs
NickCraver: As an example, if you have a large profiling tree then the serialization at the end will pay a very heavy toll in duplication...for no reason in the situation. We'd just be generating `n` copies of structures as part of a normal operation that has no racing at all.

What the lock is trying to protect is the actual reference race, but there's a premium price on the member access within that reference as a result.

src/MiniProfiler.Shared/Timing.cs
NickCraver: These will add overhead. As an example we run this for *every single request* at Stack Overflow. So the profiler can't be in the way or causing contention. As an example, we have mobile feed threads that may returns tens of thousands of tasks in a very narrow window. - Some of them will wait on the locks.

Equally important, anyone accessing these gets a copies. This really is a no-go under any circumstances. One of the founding principles of **Mini**Profiler is the lightweight nature. The amount of allocations involved for operational overhead here is simply untenable. We'll have to find another approach.

src/MiniProfiler/ProfileProviders/WebRequestProfilerProvider.cs
NickCraver: Perhaps, the name's very unintuitive (e.g. "is this only to profile async things?") but I agree that may be the sane migration approach for users.

tests/MiniProfiler.Tests/App.config
NickCraver: I'm already doing this today with the tests that have different storages, etc. required. At a higher level: it doesn't matter if it's hard to do, it matters if it's *the right thing* to do. Disabling parallelization doesn't solve the problem, it masks a problem.

tests/MiniProfiler.Tests/BaseTest.cs
NickCraver: This was explicit because, as an example, strings are `IEnumerable`. This would need to change back to the previous behavior.

src/MiniProfiler.Shared/ProfileProviders/BaseProfilerProvider.cs
NickCraver: The timing is specific to the operation, so yes, possibly making the head AsyncLocal makes the most sense. We'll have to think through if that actually breaks anything though. And the context fetch isn't free either for all requests, so it's not a clear win on the performance side.

src/MiniProfiler.Shared/Timing.cs
alexsku: I doubt the overhead of the duplication would be that high. The current code is using List<Timing> to keep children and while children are added it is doing similar duplication internally potentially multiple time. This would be just one more duplication. And it is extremely fast (less than a nanosecond for small lists (less than 20 elements) in my tests)

src/MiniProfiler.Shared/Timing.cs
alexsku: In my opinion the concurrency issue that i'm trying to fix is highly likely with web applications using async actions. Here is my pr (on top of V4) that reproduces the concurrency issue in a web app: https://github.com/alexsku/dotnet/pull/4

In the code I follow a standard pattern where the action executes multiple async service calls and then awaits for them together using Task.WhenAll. The service calls internally make an async call with ConfigureAwait(false). The completions of those calls would execute in parallel causing the issue.

Also the overhead of the lock is very small when the lock is not contested. I wrote a small test to measure the overhead: https://github.com/alexsku/dotnet/pull/5 and the overhead of lock is negligible (~25 nanoseconds). I think even with the lock the MiniProfiler stays fairly lightweight.

tests/MiniProfiler.Tests/BaseTest.cs
alexsku: Right, i didn't think about string :)

alexsku:
I found an issue describing a similar problem to which i'm trying to fix: https://github.com/MiniProfiler/dotnet/issues/35
I updated the issue pointing to this PR as a possible fix

NickCraver:
I'm closing this out as-is for all the comments above. The approach, unfortunately, just doesn't scale. If we used this at Stack Overflow we'd encounter an unsustainable increase in allocations. I promise to look at the issue in v4, but this methodology on the whole just isn't tenable.

PR #151, opened 2017-03-02
Incredibly Hacky Solution: Internal .RequestState Courier

NickCraver:
First: **THIS IS A HACK**. It's a terrible, dirty, ugly hack. The point of this PR is to find a better way.

The basic problem here is accessing information from work already done in middleware down in a view helper (form can be whatever works) when rendering a `<script>` tag for MiniProfiler.

In ASP.NET days of old, the fact that the request was buffered like crazy actually meant you could add the request profile headers *very* late in the pipe. That's not the case anymore, since headers are likely long-gone from the server when the end of your views are being rendered to the stream. The optimization is a good thing, but it means we need to re-order when some bits are done in MiniProfiler.

Since we need to calculate the unviewed MiniProfiler IDs in Middleware up front (before headers are closed for business), it's best to do it *once*, not running the storage access (which may be expensive in the remote I/O case) twice (once again for the script). The same applies for authorization. This is *likely* cheap (but up to the user) and again: only once desired. 

*Note: an example of unviewed IDs would be a POST + redirect + GET, you see both MiniProfilers on the 2nd GET there*

So middleware needs to calculate a `List<Guid>` and a `bool` and somehow get it accessible later in the view, preferably in a intuitive and elegant way. In the old days, it was this:
```c#
@MiniProfiler.RenderIncludes()
```
...which renders the `<script>` tag out for MiniProfiler. It has many `data-` attributes containing things like the `List<Guid>` above.

The new method is:
```c#
@MiniProfiler.Current.RenderIncludes()
```
It's an extension method now because `MiniProfiler.Current` spans both ASP.NET (4 & 5+) worlds and can't reference both sets of `HttpContext`, or even `HtmlString` for that matter. 

Problems with the `@` view piece:
1. In ASP.NET Core, there's no cheap way to access `HttpContext.Request` (needed to run the authorize method)
2. If I pass in a razor page to the helper (via `this`): 1) that's not guaranteed to be right if in a partial (as I understand it), and 2) that may break with RazorPages coming later.

So we have a courier in this commit, passing these values down. Things we're trying to solve:
1. Run the potentially expensive accessors once (which **have** to be in middleware before headers).
2. Make the rendering method simple/easy to use.
3. We're not adding yet another `AsyncLocal<T>` for state transitions...since we only need it downstream and not a fork up. The cost wins here (vs. running accessors again) would scale the more `async` that happens in a request.
4. It's `internal` because `MiniProfiler.Current` is often used by consumers, so we don't want extra bits hanging off there...it's bound to cause issues.

#### Current approach cons
1. It's ugly as sin.
2. There are likely corner cases on serialization if someone went crazy (hence the `internal` hiding shenanigans).
3. Internals have to be exposed *at all* downstream.
4. I had trouble sleeping after writing this.

#### Other options considered (and problems)
1. Tag helpers (with context access):
 - Pro: Very clean, potentially `<script miniprofiler>` or something
 - Con: The user has to register tag helpers (a library can't do this as far as I'm aware)
 - Con: Authorization/Unviewed accessors would still run twice without some storage
 - ??: I'm not sure you can ever access context anyway. Still, this would be nice with a courier, if we can register them automatically.
2. `IHttpContextAccessor`:
 - Pro: it works, much like older MiniProfiler
 - Con: Expensive, there's a reason it's off by default
 - Con: Accessors still running twice
3. Add a `AsyncLocal<RequestState>`
 - Pro: Fairly clean
 - Con: Adds more `async` context switch actions...which would add up
 - Con: lots of unnecessary forking. We just want it available, not forked...so it's very wasteful

Does anyone have a better approach here? Surely something exists that's better suited and relatively inexpensive.

cc @DamianEdwards @davidfowl @benaadams


hiiru:
One other option could be an IHtmlHelper extension.
It's similar to this approach, but without the internals stuff and provides access to the HttpContext.

Here is a Quick'n'Dirty example:
https://github.com/hiiru/dotnet/tree/v4-HtmlHelper

However somehow it just feels like a wrapper around an extension method... meh.

src/MiniProfiler.AspNetCore/RequestState.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// Stores the request state, passed down in the <see cref="HttpContext"/>
+    /// </summary>
+    internal class RequestState
+    {
+        internal const string HttpContextKey = "__MiniProfiler.RequestState";
+
+        /// <summary>
+        /// Is the user authorized to see this MiniProfiler?
+        /// </summary>
+        public bool IsAuthroized { get; set; }
```
juunas11: Is this an intentional typo? :) Shouldn't it be `IsAuthorized`?

src/MiniProfiler.AspNetCore/MiniProfilerMiddleware.cs
```diff
@@ -100,12 +100,15 @@ public class MiniProfilerMiddleware
             }
         }
 
-        private async Task AppendHeadersAsync(HttpContext context, MiniProfiler current)
+        private async Task SetHeadersAndState(HttpContext context, MiniProfiler current)
         {
             try
             {
+                // Are we authorized???
+                var isAuthroized = Options.ResultsAuthorize?.Invoke(context.Request) ?? true;
```
juunas11: Shouldn't this also be `isAuthorized`? :)

NickCraver: Yep...forgot to fix these, done!

PR #155, opened 2017-03-19
Remove unnecessary `Cast<string>()`

justinvp:
`form.Keys` is already `ICollection<string>`, there's no need to cast.


NickCraver:
Old code from old .NET indeed, thanks!

PR #157, opened 2017-03-19
Remove explicit static constructors

justinvp:
Explicit static cctors cause the C# compiler to not mark types as beforefieldinit, which means the JIT will add checks to each static method and instance constructors of the type to make sure that the static constructor was previously called. This can be avoided by removing explicit static constructors and initializing static fields inline.

See https://msdn.microsoft.com/en-us/library/ms182275.aspx for more info.


NickCraver:
Doh, I never did fix these last 2 when I got sidetracked on how to unify settings best between both libs - thanks!

PR #159, opened 2017-03-19
Use StringBuilder.Append(char) for single chars

justinvp:
Slightly cheaper and in some cases slightly simpler code (e.g. `'"'` vs. `"\""`). Though, I wouldn't expect much of a measurable difference in these cases.


NickCraver:
Agree on more readable as well - merging!

Issue #144, opened 2016-11-30
MiniProfiler v4


NickCraver:
I'm working on v4 of MiniProfiler as time allows, but an overall progress checklist would probably be helpful. If anyone has comments, suggestions, reminders of breaking changes we should make now, etc. - please, chime in.

You can track progress ~~[on the v4 branch here](https://github.com/MiniProfiler/dotnet/tree/v4)~~ now [in the masterbranch]: (https://github.com/MiniProfiler/dotnet). V3 has been branched off for any servicing [here](https://github.com/MiniProfiler/dotnet/tree/v3).

The library layout I currently have splits the primary libraries into `MiniProfiler` (current lib), `MiniProfiler.AspNetCore`, and `MiniProfiler.Shared` (referenced by both).

The storage for things like SQL Server are in `MiniProfiler.Providers.SqlServer`, and even `MiniProfiler.Providers.SqlServerCe`.  Now the SQL Server *profilers* are free (all ADO.NET providers are, being based on `DbConnection`), so it's essentially free to include "in the box" in `MiniProfiler.Shared` with no additional references. So I maintain it there and make life easier. The SQL Server libs are only needed for storage.

In other cases like `MiniProfiler.Providers.RavenDB`, `MiniProfiler.Providers.MongoDB`, etc. they'll have both a profiler *and* a storage provider on the package. I think this makes the most sense of the available options and tradeoffs in reference weight. That means providers and platforms that support various `netstandard` levels can just work on the most things possible.

New major features:
- [ ] [Server Timing Headers](http://wicg.github.io/server-timing/)
  - **Blocked**: ASP.NET Core [lacks (realistic) support for `Trailer` headers](https://github.com/aspnet/KestrelHttpServer/issues/622)
  - Methods are in to generate it (based on custom timings), trailers are the issue.
- [x] ASP.NET Core support
  - [x] `<mini-profiler>` tag helper
  - [x] `<profile name="My Step">` tag helper
- [x] Async support
- [x] Multi-threaded access support (goes with `async`)

Project System:
- [x] Convert to new .csproj (VS 2017)
- [x] Port tests to XUnit
- [x] Get all tests passing
- [x] Get `.nupkg`s building with new .csproj

Target Frameworks:
- [x] `netstandard1.5` (for all non-`System.Web`)
- [x] `net46` (for all)

API Changes (**breaking**):
- [x] Adding `Name` to SQL Storage providers (#104)
[...]

PR #165, opened 2017-03-29
Redis Storage for v4

jdaigle:
Basic implementation using [StackExchange.Redis](https://www.nuget.org/packages/StackExchange.Redis.StrongName/) with the naming convention suggested here: https://github.com/MiniProfiler/dotnet/issues/144#issuecomment-280232091.

Architecturally, the results are stored across three data structures:

* JSON serialized results are stored as strings using the profiler Id (and a configurable prefix) as the key. Keys expire based on configurable duration.
* A list of profiler Ids (using a configurable key). New results are inserted at the head. The list is trimmed to a configurable limit.
* A set of "unviewed" profiler Ids per user using the user (and a configurable prefix) as the key.

The only issue I've run into with this design is that it doesn't easily support the `start` and `finish` parameters for `List()`. However, at least in terms of code in this repository, there are no consumers that specify those parameters.


src/MiniProfiler.Providers.StackExchange.Redis/RedisStorage.cs
```diff
@@ -0,0 +1,263 @@
...
+            DateTime? finish = null,
+            ListResultsOrder orderBy = ListResultsOrder.Descending)
+        {
+            // TODO: start/finish not yet supported - will simply return all results
+
+            IEnumerable<RedisValue> ids;
+            switch (orderBy)
+            {
+                case ListResultsOrder.Ascending:
+                    ids = (await _database.ListRangeAsync(ProfilerResultListKey, maxResults + 1, -1).ConfigureAwait(false)).Reverse();
```
NickCraver: I *think* you can pass parameters in backwards here or some such - @mgravell?

src/MiniProfiler.Providers.StackExchange.Redis/RedisStorage.cs
```diff
@@ -0,0 +1,263 @@
...
+        /// <returns>The loaded <see cref="MiniProfiler"/>.</returns>
+        /// <remarks>
+        /// Should also update that the resulting profiler has been marked as viewed by its
+        /// profiling <see cref="MiniProfiler.User"/>.
+        /// </remarks>
+        public async Task<MiniProfiler> LoadAsync(Guid id)
+        {
+            RedisKey key = ProfilerResultKeyPrefix.Append(id.ToString());
+            RedisValue value = await _database.StringGetAsync(key).ConfigureAwait(false); ;
+            return MiniProfiler.FromJson(value);
```
NickCraver: We can use `protobuf-net` for protocol buffers encoding here and save a ton of string allocations (we're doing this at Stack Overflow today, it works already because of `[DataMember]` attributes on `MiniProfiler` already)...I can help with sample code there!

src/MiniProfiler.Providers.StackExchange.Redis/RedisStorage.cs
```diff
@@ -0,0 +1,263 @@
...
+        /// </summary>
+        /// <param name="profiler">The <see cref="MiniProfiler"/> to save.</param>
+        /// <remarks>
+        /// Should also ensure the profiler is stored as being un-viewed by its profiling <see cref="MiniProfiler.User"/>.
+        /// </remarks>
+        public async Task SaveAsync(MiniProfiler profiler)
+        {
+            var id = profiler.Id.ToString();
+            RedisKey key = ProfilerResultKeyPrefix.Append(id);
+            RedisValue value = profiler.ToJson();
```
NickCraver: See below comment on Protobuf serialization (@mgravell is this build into SE.Redis anywhere at this point?)

src/MiniProfiler.Providers.StackExchange.Redis/RedisStorage.cs
```diff
@@ -0,0 +1,263 @@
...
+        /// <summary>
+        /// Asynchronously sets a particular profiler session to "viewed"
+        /// </summary>
+        /// <param name="user">The user to set this profiler ID as viewed for.</param>
+        /// <param name="id">The profiler ID to set viewed.</param>
+        public async Task SetViewedAsync(string user, Guid id)
+        {
+            RedisKey key = ProfilerResultUnviewedSetKeyPrefix.Append(user);
+            RedisValue value = id.ToString();
+            await _database.SetRemoveAsync(key, value).ConfigureAwait(false);
```
NickCraver: Can remove `async`/`await` and just return the call as your `Task`

src/MiniProfiler.Providers.StackExchange.Redis/RedisStorage.cs
```diff
@@ -0,0 +1,263 @@
...
+        /// <returns>The loaded <see cref="MiniProfiler"/>.</returns>
+        /// <remarks>
+        /// Should also update that the resulting profiler has been marked as viewed by its
+        /// profiling <see cref="MiniProfiler.User"/>.
+        /// </remarks>
+        public MiniProfiler Load(Guid id)
+        {
+            RedisKey key = ProfilerResultKeyPrefix.Append(id.ToString());
+            RedisValue value = _database.StringGet(key);
+            return MiniProfiler.FromJson(value);
```
NickCraver: See below comments on Protobuf

src/MiniProfiler.Providers.StackExchange.Redis/RedisStorage.cs
jdaigle: I'm not necessarily opposed to taking on protobuf-net as a dependency. But maybe instead we could allow users to customize the serialization - either an interface or just a pair of delegates - to avoid the hard dependency. I assume it could a pretty straightforward call.

e.g. `RedisValue Encode(MiniProfiler profiler);`

src/MiniProfiler.Providers.StackExchange.Redis/RedisStorage.cs
jdaigle: Is this regarding `Reverse()`? I think Redis will always return list ranges in list order - and so I've tried to optimized for what I think is the usual case (i.e. descending).

src/MiniProfiler.Providers.StackExchange.Redis/RedisStorage.cs
```diff
@@ -0,0 +1,263 @@
...
+        /// <param name="start">(Optional) The start of the date range to fetch.</param>
+        /// <param name="finish">(Optional) The end of the date range to fetch.</param>
+        /// <param name="orderBy">(Optional) The order to fetch profiler IDs in.</param>
+        public IEnumerable<Guid> List(
+            int maxResults,
+            DateTime? start = null,
+            DateTime? finish = null,
+            ListResultsOrder orderBy = ListResultsOrder.Descending)
+        {
+            // TODO: start/finish not yet supported - will simply return all results
```
jdaigle: I did consider using a sorted set, instead of a list, and using the prolifer's timestamp as the score. Then I can query with ZRANGEBYSCORE. But I was ever so slightly worried that write performance would suffer, especially since List is comparatively less used (this is just a bold assumption, I don't know how others use it).

But maybe the performance of sorted sets isn't worth worrying about. Thoughts?

src/MiniProfiler.Providers.StackExchange.Redis/RedisStorage.cs
NickCraver: I'm not so concerned with the dependency, but would hate for that layer to be *in the provider* - it would have to be some interface in the base .Shared library for serializing, but then you've got the problem of needing the user to have `n` library additions (e.g. another lib which *does* require protobuf) just for the serialization piece, so they're included `MiniProfiler.*`, `MiniProfiler.Providers.StackExchange.Redis`, and `<something>` which uses protobuf. IMO, that's too much, and pretty confusing.

Now we are changing up `StackExchange.Redis` to have a L1/L2 caching layer (currently in another library internally at Stack), so everyone can use it. I'd vote we can *remove* `protobuf` there if needed later, and match *that* serialization story. After all, the serialization target constraints are per-environment. For example most document stores are JSON, redis is bytes, relational is typed columns etc.

IMO, a `protobuf-net` dependency is one of the fastest, slimmest options and the simplest thing for a user to get setup.

src/MiniProfiler.Providers.StackExchange.Redis/RedisStorage.cs
NickCraver: If we're only keeping ~100, I think we're good with a sorted set. We've had this *exact same talk* about how Stack Overflow samples are stored...but I can't remember exactly what we concluded, @mgravell?

NickCraver:
Still on track to merge this and fix issues btw, I just haven't had time to complete the other blocker with EF first. We'll likely have to rebase the `.sln`, but the rest should go clean. I'm catching up on several dozen issues across projects then trying to get wrapped that up.

jdaigle:
Updates:

* Rebased with master, so the merge should be clean now.
* Refactored the results list to use a sorted set - that ended up being pretty clean and straight forward.
* I haven't yet incorporated `protobuf-net`, but I'm sold on your arguments.

Unfortunately I have been running into some issues (see #166) that have been slowing me down.

NickCraver:
Thanks @jdaigle, sorry for the busy week - just catching up now. All your work is greatly appreciated!

jdaigle:
@NickCraver thanks!.

PR #167, opened 2017-04-02
Add support for profiling ASP.NET Web Services (.asmx) AJAX calls

jdaigle:


NickCraver:
Ah I forgot to bring some of these back - thanks! As an aside, definitely need to get rid of that `eval()` fallback all over, I'll replace it with JSON define if missing.

PR #168, opened 2017-04-02
Wrap fetch for profiler

benaadams:
Wrap the client side fetch api to get profiler measurements when that is used rather than XHR

fetch


NickCraver:
Awesome, thanks!

PR #170, opened 2017-04-03
Use cached StringBuilder rather than StringBuilder.Replace

benaadams:
Bit uglier but sb.Replace was sky high on the execution

Before
before

After
after

### DIFF
--- a/src/MiniProfiler.AspNetCore/MiniProfilerExtensions.cs
+++ b/src/MiniProfiler.AspNetCore/MiniProfilerExtensions.cs
@@ -73,15 +73,10 @@ namespace StackExchange.Profiling
             var ids = state?.RequestIDs;
             if (ids != null)
             {
-                var first = true;
                 var length = ids.Count;
                 for (var i = 0; i < length; i++)
                 {
-                    if (first)
-                    {
-                        first = false;
-                    }
-                    else
+                    if (i > 0)
                     {
                         sb.Append(',');
                     }
--- a/src/MiniProfiler.AspNetCore/MiniProfilerExtensions.cs
+++ b/src/MiniProfiler.AspNetCore/MiniProfilerExtensions.cs
@@ -1,8 +1,8 @@
 ﻿using Microsoft.AspNetCore.Html;
 using Microsoft.AspNetCore.Http;
-using StackExchange.Profiling.Helpers;
 using System;
-using System.Linq;
+using System.Globalization;
+using System.Runtime.CompilerServices;
 using System.Text;
 
 namespace StackExchange.Profiling
@@ -14,6 +14,19 @@ namespace StackExchange.Profiling
     {
         // don't allocate this every time...
         private static readonly HtmlString includeNotFound = new HtmlString("<!-- Could not find 'include.partial.html' -->");
+
+        [ThreadStatic]
+        private static StringBuilder _stringBuilderCache;
+
+        private static StringBuilder GetStringBuilder() => _stringBuilderCache ?? CreateStringBuilder();
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private static StringBuilder CreateStringBuilder()
+        {
+            var sb = new StringBuilder();
+            _stringBuilderCache = sb;
+            return sb;
+        }
 
         /// <summary>
         /// Renders script tag found in "include.partial.html".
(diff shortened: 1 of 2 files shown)


benaadams:
Also looking at the call count; it seems a bit high compared to the layout view

I'm not to knowledgable about taghelpers, but does the wildcard mean its run for every tag?
```@addTagHelper "*, MiniProfiler.AspNetCore.Mvc"```

e.g. could that be more specific? /cc @DamianEdwards

**Comment by NickCraver** [comment]
There's a problem here in that it removes the ability to template in general - I already question how much this is used out in the world, because there's certainly a tradeoff between templating, performance, and code duplication any way you slice it.

I see a few options:
1. Drop templating support in v4
2. Fast path no template path being set (duplicating in C# what's in the default template, which only serves as an example after doing so)
3. Accept terrible performance for life (this isn't an option, just including it)
4. Somehow add templating support via another lib? If there was demand maybe? This is a variant of 2.

Given the size of the strings in play in a few cases, I'd say we switch to a `StringBuilder` and build it up directly in code - no matter how ugly it is. Thoughts? There's another code path for `share.html` (much more rarely accessed) that's similarly optimized for sanity over performance at the moment. They should get the same love whichever way we go, I think.

Thoughts?

**Comment by benaadams** [comment]
Since its a light weight profiler; I think would prefer to take the uglier `[ThreadStatic] StringBuilder` and then revisit with a more flexible solution later?

Was trying to workout what the slow bit it was telling me was in may page and it turned out to be the profiler tag 😢

**Comment by benaadams** [comment]
@NickCraver new and improved?

File: src/MiniProfiler.AspNetCore/MiniProfilerExtensions.cs [hunk 1]
```diff
@@ -40,33 +53,76 @@ public static class MiniProfilerExtensions
...
+            sb.Append("\" data-ids=\"");
+            var ids = state?.RequestIDs;
+            if (ids != null)
             {
-                return includeNotFound;
+                var first = true;
+                var length = ids.Count;
+                for (var i = 0; i < length; i++)
+                {
+                    if (first)
```
NickCraver: if (i > 0) simplification?

src/MiniProfiler.AspNetCore/MiniProfilerExtensions.cs
```diff
@@ -40,33 +53,76 @@ public static class MiniProfilerExtensions
...
-              .Replace("{showChildren}", toJs(showTimeWithChildren ?? MiniProfiler.Settings.PopupShowTimeWithChildren))
-              .Replace("{maxTracesToShow}", (maxTracesToShow ?? MiniProfiler.Settings.PopupMaxTracesToShow).ToString())
-              .Replace("{showControls}", toJs(showControls ?? MiniProfiler.Settings.ShowControls))
-              .Replace("{authorized}", toJs(authorized))
-              .Replace("{toggleShortcut}", MiniProfiler.Settings.PopupToggleKeyboardShortcut)
-              .Replace("{startHidden}", toJs(startHidden ?? MiniProfiler.Settings.PopupStartHidden))
-              .Replace("{trivialMilliseconds}", MiniProfiler.Settings.TrivialDurationThresholdMilliseconds.ToString());
-            return new HtmlString(sb.ToString());
+            sb.Append("\" data-position=\"");
+            sb.Append((position ?? MiniProfiler.Settings.PopupRenderPosition).ToString().ToLower());
```
NickCraver: After this commit I have an urge to `switch` that `.ToString().ToLower()`. Damn you, instigator.

src/MiniProfiler.AspNetCore/MiniProfilerExtensions.cs
```diff
@@ -40,33 +53,76 @@ public static class MiniProfilerExtensions
...
+
+            sb.Append("\" data-trivial=\"");
+            sb.Append((showTrivial ?? MiniProfiler.Settings.PopupShowTrivial) ? "true" : "false");
+
+            sb.Append("\" data-children=\"");
+            sb.Append((showTimeWithChildren ?? MiniProfiler.Settings.PopupShowTimeWithChildren) ? "true" : "false");
```
NickCraver: +1 - I'll look at the JS, perhaps we can `if` these for existence there instead. It'll need a change in both libraries, but doable if we're ditching templates. I'll handle the .Shared move after this is merged in.

src/MiniProfiler.AspNetCore/MiniProfilerExtensions.cs
```diff
@@ -15,6 +15,19 @@ public static class MiniProfilerExtensions
...
+
+        private static StringBuilder GetStringBuilder() => _stringBuilderCache ?? CreateStringBuilder();
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private static StringBuilder CreateStringBuilder()
+        {
+            var sb = new StringBuilder();
+            _stringBuilderCache = sb;
+            return sb;
+        }
```
NickCraver: I agree with the approach (and happy to merge as-is). I'll likely share this elsewhere though, thoughts on me bringing https://github.com/opserver/Opserver/blob/overhaul/Opserver.Core/StringBuilderCache.cs into `MiniProfiler.Shared\Internal` after this is merged in?

src/MiniProfiler.AspNetCore/MiniProfilerMiddleware.cs
```diff
@@ -50,7 +51,8 @@ public class MiniProfilerMiddleware
             if (basePath.StartsWith("~/", StringComparison.Ordinal)) basePath = basePath.Substring(1);
             if (basePath.EndsWith("/", StringComparison.Ordinal) && basePath.Length > 2) basePath = basePath.Substring(0, basePath.Length - 1);
 
-            BasePath = new PathString(basePath);
+            MatchPath = new PathString(basePath);
+            BasePath = new PathString(basePath.EnsureTrailingSlash());
```
NickCraver: This may break the share page include or cause a double `//`, in it's current form. Will need to check and sync behavior with both libs.

benaadams: Was trying to do it up front in `.ctor` and avoid the RegEx per call. Happy to take direction

src/MiniProfiler.AspNetCore/MiniProfilerExtensions.cs
benaadams: Yeah, that looks better

src/MiniProfiler.AspNetCore/MiniProfilerExtensions.cs
benaadams: Let me change this, for pride 😉

benaadams:
Most of the heavy parts are now external; MemoryCache which was fixed by https://github.com/aspnet/Caching/pull/246 and Guid which will see if can take a look at

Enviorment.MachineName is expensive used by `MinProfiler.ctor`; does it need to be looked up each time? vs static

benaadams:
Good with merge and then iterate on next bits?

src/MiniProfiler.AspNetCore/MiniProfilerMiddleware.cs
NickCraver: I totally agree with your change, just need to make the other lib match when removing all templating...more of a note for me there.

NickCraver:
Agreed - let me merge in and fix up the impacts in review, also bringing in StringBuilder cache. This will remove templating in v4 but I'm okay with that. If many want it back, we can do so in a more efficient way.

benaadams:
Added PR for Guid.ToString https://github.com/dotnet/coreclr/pull/10669

NickCraver:
Looking forward to the `.ToString()` changes there.

FYI, some followup commits: e88ace59e08aa4b3db10b35d5b78e79968bd6374 and 2b9b5aa038c75753c22581dc8b86de6fc34c8f48 to simplify and DRY things. Breaking change of removing templating for V4 is complete, I'll add it to notes in #144

PR #172, opened 2017-04-11
Support ignoring request paths matching after a file extensions

### LINKED ISSUE #171: Cannot ignore paths that must match path info after a "file extension"
ASP.NET Web Services can expose JavaScript resources through paths such as `/MyWebService.asmx/js`. I want to be able to profile requests for operations on that web service (e.g. `/MyWebService.asmx/Foo`) but I want to ignore the dynamic JavaScript resources.

I want to be able to include `".asmx/js"` in `MiniProfilerWebSettings.IgnoredPaths`, but this doesn't work. `WebRequestProfilerProvider` only compares ignored paths against `HttpRequest. AppRelativeCurrentExecutionFilePath`, which only contains the path up through the "file extension".

It should work to change the logic to compare against `HttpRequest.Path`, but this is a breaking change (particularly for anyone that hosts from a virtual directory). Thoughts?

jdaigle:
Fixes #171 

Implementing a slightly better, albeit relatively naive, GetAppRelativePath() which will actually include all of the path information (minus query string) after any "file extension".

While this causes in an string allocation, it's no worse than what `AppRelativeCurrentExecutionFilePath` is doing under the hood.

Also, a little bit of cleanup such as accessing Request.Path just once, and accessing Request.Url only when we need it.

### DIFF
--- a/src/MiniProfiler/ProfileProviders/WebRequestProfilerProvider.cs
+++ b/src/MiniProfiler/ProfileProviders/WebRequestProfilerProvider.cs
@@ -30,27 +30,27 @@ namespace StackExchange.Profiling
         public override MiniProfiler Start(string profilerName = null)
         {
             var context = HttpContext.Current;
-            if (context?.Request.AppRelativeCurrentExecutionFilePath == null) return null;
+            var path = context?.Request.Path;
+            if (path == null) return null;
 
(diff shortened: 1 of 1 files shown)


NickCraver:
Looks sane - will take a closer look locally, thanks!

PR #174, opened 2017-04-15
Fix (hack?) for AsyncLocal<T> Head and ASP.NET Pipeline Events

### LINKED ISSUE #173: Issues with AsyncLocal<T> and ASP.NET on .NET 4.6.2
Framework version 4.6.2

For certain HTTP requests anything set to an `AsyncLocal<T>` variable in `Application_BeginRequest()` is null by the time any part of the MVC5 pipeline invokes (action filters, controllers, etc). Further, the value is still null at `Application_EndRequest()`. On my computer, this is easily reproducible from an empty ASP.NET project using an HTTP POST with a *relatively large* content body, 14KB or so).

I know this isn't exactly an issue with MiniProfiler, but maybe it is now that MiniProfiler uses `AsyncLocal<T>`.

For example, [`IDbProfiler.ExecuteFinish()`](https://github.com/MiniProfiler/dotnet/blob/master/src/MiniProfiler.Shared/MiniProfiler.IDbProfiler.cs) may throw a null reference exception the timing object added to the dictionary in `IDbProfiler.ExecuteStart()` is null (because it tries to access `Head` via `AsyncLocal<>` which is sometimes null).

Side note: how would you even go about reporting this type of .NET bug (if it is a bug)? Maybe we [can't rely on `AsyncLocal` in ASP.NET (not Core)?](http://stackoverflow.com/questions/36511243/safety-of-asynclocal-in-asp-net-core).

jdaigle:
Fixes #173 

There is an issue in ASP.NET when the profiler is started in BeginRequest, but
accessed in an HttpHandler later on, in which `Head` is null (it should not be
null until the profiler is stopped).

The `Head` property is now backed by `AsyncLocal<T>` which is tied to the current
thread's `ExecutionContext`. However the `ExecutionContext` does not flow between
ASP.NET pipeline events if the events fire on separate threads.

This fix (a hack really) ensures that `Head` is not null by setting it to `Root`.
In practice I think this will be OK since it's unlikely a user will create a timing,
other than root, which spans pipeline events.

### DIFF
--- a/src/MiniProfiler/ProfileProviders/WebRequestProfilerProvider.cs
+++ b/src/MiniProfiler/ProfileProviders/WebRequestProfilerProvider.cs
@@ -222,7 +222,20 @@ namespace StackExchange.Profiling
         /// </summary>
         private MiniProfiler Current
         {
-            get => HttpContext.Current?.Items[CacheKey] as MiniProfiler;
+            get
+            {
+                var profiler = HttpContext.Current?.Items[CacheKey] as MiniProfiler;
+                if (profiler != null && profiler.Head == null)
+                {
+                    // 'Head' is backed by 'AsyncLocal<T>', which is attached to the ExecutionContext,
+                    // and doesn't flow between ASP.NET pipeline events (BeginRequest/etc.) if the
+                    // events fire on separate threads.
+                    // This is a hack to ensure the Head is not null after the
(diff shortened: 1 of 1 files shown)


jdaigle:
Closing since we've implemented an alternative fix: https://github.com/MiniProfiler/dotnet/pull/177

PR #176, opened 2017-04-30
Entity Framework 6 Simplification - fix for #175

### LINKED ISSUE #175: MiniProfiler.EF6 incompatible with custom DbProviderServices implementation
I'm currently trying out MiniProfiler.AspNetCore.Mvc with MiniProfiler.EF6. Both on v4.0.0-alpha6-52.

When configuring MiniProfiler.EF6 I run into an issue. We use a custom DbProviderServices implementation (CachingProviderServices from the EFCache project), which makes the constructor of EFProfiledDbProviderServices throw.

It seems the constructor looks for the Instance property through reflection, which doesn't exist in the DbProviderServices class, but does exist in the default implementation used by EF (SqlProviderServices).

The following snippet throws the exception:
`Unable to define EFProfiledDbProviderServices class of type 'CachingProviderServices'. Please check that your web.config defines a <DbProviderFactories> section underneath <system.data>.`
``` c#
        public static void Main(string[] args)
        {
            // Setup EFCache  
            DbConfiguration.Loaded +=
                (sender, args) => args.ReplaceService<DbProviderServices>(
                    (dbServices, _) => new CachingProviderServices(dbServices, transactionHandler, cachingPolicy));
            DbInterception.Add(transactionHandler);

            // Setup MiniProfiler.EF6
            MiniProfilerEF6.Initialize();

            // This line throws the exception.
            new DbContext(GetConnectionString())
                .Initialize(false);
        }
```

A workaround could be implemented by creating an overload for MiniProfilerEF6.Initialize() which accepts the SqlProviderServices instance. The user can call this overload from the DbConfiguration.Loaded event, just like CachingProviderServices in the sample above.

Of course this makes it harder to use than the current call to Initialize (which is probably sufficient for 90% of the users?), but those users can of course just use the no parameters variant, which already exists.

NickCraver:
Previously we were relying on reflection for the .Instance static access on the type. But this is realistic scenario - you need to wrap *another* provider to do anything useful (ya know, profiling). So we need not be concerned with this access pattern and can be faster and support more without it.

This should resolve #175 and be faster and simpler for everyone.

@jdaigle you have more experience with EF6 than I by far - can you eyeball this changeset please? I did some digging and it looks like the roots of what was there was back in EF 6 alpha 3, so it hasn't gotten love in a very long time.


jdaigle:
I don't see anything initially that stands out as a problem. And it's definitely a lot simpler and easier to grok. In the next day or so I should be able to test this code with some of our more "exotic" EF projects, but I suspect it'll work fine.

NickCraver:
Merging in so we can test out this simplified version

PR #182, opened 2017-05-09
GetServerTimingHeader should use milliseconds instead of seconds

karl-sjogren:
GetServerTimingHeader should use milliseconds instead of seconds according to the [W3C specification](https://w3c.github.io/server-timing/#the-code-dfn-performanceservertiming-dfn-code-interface) ( [DOMHighResTimeStamp ](https://www.w3.org/TR/hr-time-2/#idl-def-domhighrestimestamp) to be exact).

This was a bug in the initial implementation in Chrome but was fixed in [Issue 704340](https://bugs.chromium.org/p/chromium/issues/detail?id=704340) and is in the current beta branch.

I also added proper NumberFormatInfo to the calls that turn the numbers into strings to account for different culture settings.

### DIFF
--- a/src/MiniProfiler.Shared/MiniProfiler.ServerTiming.cs
+++ b/src/MiniProfiler.Shared/MiniProfiler.ServerTiming.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Generic;
+using System.Globalization;
 using System.Text;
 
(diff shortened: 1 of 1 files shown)


NickCraver:
Indeed! Great write up, links, and fix - thank you!

PR #188, opened 2017-06-13
Add MySQL Server async storage backend

bgrainger:
This adds an `IAsyncStorage` implementation that stores data in MySQL Server. It uses my [MySqlConnector library](https://github.com/mysql-net/MySqlConnector) so DB operations are truly async. It builds for `net451` and `netstandard1.5`.

Q: Do you want this code as part of the core MiniProfiler repository, or would you prefer it to be in its own repo and maintained separately?

90% of the code was ["derived from"](http://typicalprogrammer.com/introduction-to-abject-oriented-programming) `SqlServerStorage.cs`; if you are OK with the code being in this repo, I could refactor out that duplicated code into a new `StandardDatabaseStorageBase` abstract base class that contains the core DB logic and gets the SQL statements from its derived classes. (That could be done in this PR or a follow-up one.)

### DIFF
--- a/MiniProfiler.sln
+++ b/MiniProfiler.sln
@@ -63,6 +63,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "benchmarks", "benchmarks", 
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MiniProfiler.Benchmarks", "benchmarks\MiniProfiler.Benchmarks\MiniProfiler.Benchmarks.csproj", "{9C653E9B-71DA-49D3-83BE-ED78873DE247}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MiniProfiler.Providers.MySql", "src\MiniProfiler.Providers.MySql\MiniProfiler.Providers.MySql.csproj", "{36ED231C-19C9-4CA3-A085-4066A9B571CD}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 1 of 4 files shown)


NickCraver:
@bgrainger nice! I'm not worried about the de-dupe yet since these are so static, I'd say let's worry about that after release even. I've heard someone ask for Postgres and someone mention Oracle storage as well. If those come along, what's in a base may be very different anyhow.

This is much appreciated, pulling into the main repo :)

NickCraver:
@bgrainger What's the intended build targets here? The csproj has net451 and checks against net45 and net46 - the target frameworks get mixed at the last minute, or something I'm not seeing here?

bgrainger:
> The csproj has net451 and checks against net45 and net46 - the target frameworks get mixed at the last minute, or something I'm not seeing here?

Copy/paste error from the SqlServer project 😳 thanks for catching that.

It should be `net451` (MySqlConnector minimum framework) and `netstandard1.5` (MiniProfiler.Shared minimum); will push a fix.

BTW, I noticed that MiniProfiler.Providers.SqlServer/SqlServerStorage.cs targets both `net45` and `net46`, but doesn't have any conditional compilation logic for .NET 4.6; is there an advantage to targeting both frameworks (and should I do the same here)?

NickCraver:
@bgrainger yep we'll need a `net46` build here too, it's intentional due to `net45*` being a `AsyncLocal` shim to take the lack of it whereas `net46` and higher has it proper. I'll yank this in then tweak, thanks for the work!!

NickCraver:
@bgrainger before I can release this - we've got a mismatch...do you plan to strong-name the connector library? Since MiniProfiler is strongly named (as all our libraries have to be, unfortunately), the dependencies must be as well.

bgrainger:
I hadn't been planning to, based on my reading at the time I responded to mysql-net/MySqlConnector#224.

But I'm open to having my mind changed (after reading things like aspnet/DataProtection#245 recently)... is there a place we could chat about it (that's not a public GitHub issue)?

NickCraver:
Yep - I'm craver at stackoverflow. The basics are: some people need it, some people don't. But anyone that does has to have it all the way up the dependency tree. I personally despise strong naming, but it's way less effort to *just do it* than to argue about it. We're so tired of the debate. StackExchange.Redis, Dapper, etc. all have `.StrongName` versions, but that was a huge mistake - we're strong naming all of them in vNext and deprecating the `.StrongName` versions. Because without it, the dependency trees between strong naming and not are crazy and cause issues.

TL;DR: Strong-naming sucks. But as an author, it sucks more *not* doing it...that's why most major libraries central to the ecosystem are strong-named.

Happy to answer any questions in email, and I'll copy Marc as well. Also open to DMs on Twitter: [@Nick_Craver](https://twitter.com/Nick_Craver)

PR #189, opened 2017-06-16
Entity Framework 6 Profiling Enhancements

### LINKED ISSUE #130: MiniProfiler EF6 and batch queries
I've been trying to get MiniProfiler to play nicely with two different query batching tool's "future" query support.  Both EntityFramework.Extended and EntityFramework-Plus have methods for adding "Future()" to a query and then executing several queries together when the first one is enumerated.  This is a feature which has native support in NHibernate.  Both products work but neither is showing the batched queries in MiniProfiler's output.  With NHibernate the matched queries are displayed in MiniProfiler's output.

Best I've been able to figure out, it appears that EF is being served a non-profiled SQL connection for batch operations.  Is there a way to make EF use profiled connections for these other operations?

jdaigle:
This pull request provides a number of enhancements to profiling EF6 based on our usage.

1. Cache the `EFProfiledDbProviderServices` instance.

Internally DbProviderServices is registered as a singleton. However when the service is "replaced" via `ReplaceService<>()`, the interception func is executed *every time* the service is resolved. We can cache the `EFProfiledDbProviderServices` instance (and anything else we want to replace) based on the type and key.

This should hopefully help keep the profiling overhead low.

2. The current implementation, which replaces `EFProfiledDbProviderServices`, only supports cached commands that cloned when needed by EF. The `Clone()` method creates a new `ProfiledDbCommand` with the current `MiniProfiler`. However for some queries, EF6 will use `DbConnection.CreateCommand()` - and since we do not setup a `ProfiledDbConnection`, these queries *are not profiled*.

To fix these, we can wrap both `DbProviderFactory` and EF's `IDbConnectionFactory` - EF will use one or the other based on the type of connection string and how the DbContext is constructed.

This should also resolve #130.

Unfortunately to support this additional wrapping, we also need to implement our own `IDbProviderFactoryResolver` and a resolver for `IProviderInvariantName`. The implementation of `EFProfiledInvariantNameResolver ` in particular is **not pretty**. But it works.

Based on our initial testing, these these work without issue. Hopefully it works as well for others.

P.S. EF6 is annoying as hell. But I'm working hard to make this thing work.

### DIFF
--- a/src/MiniProfiler.EF6/EFProfiledDbProviderFactoryResolver.cs
+++ b/src/MiniProfiler.EF6/EFProfiledDbProviderFactoryResolver.cs
@@ -5,8 +5,8 @@ using StackExchange.Profiling.Data;
 namespace StackExchange.Profiling.EntityFramework6
 {
     /// <summary>
-    /// Wrapper for a service for obtaining the correct System.Data.Common.DbProviderFactory from
-    //  a given System.Data.Common.DbConnection.
+    /// Wrapper for a service for obtaining the correct <see cref="DbProviderFactory"/> from
+    /// a given <see cref="DbConnection"/>.
     /// </summary>
     public class EFProfiledDbProviderFactoryResolver : IDbProviderFactoryResolver
     {
@@ -15,15 +15,15 @@ namespace StackExchange.Profiling.EntityFramework6
         /// <summary>
         /// Initialises a new instance of the <see cref="EFProfiledDbProviderFactoryResolver"/> class.
         /// </summary>
-        /// <param name="inner">The IDbProviderFactoryResolver to wrap.</param>
+        /// <param name="inner">The <see cref="IDbProviderFactoryResolver"/> to wrap.</param>
         public EFProfiledDbProviderFactoryResolver(IDbProviderFactoryResolver inner)
         {
             _inner = inner;
         }
 
         /// <summary>
-        /// Returns the System.Data.Common.DbProviderFactory for the given connection,
-        /// unwrapping the ProfiledDbConnection as necessary
+        /// Returns the <see cref="DbProviderFactory"/> for the given connection,
+        /// unwrapping the <see cref="ProfiledDbConnection"/> as necessary
         /// </summary>
         /// <param name="connection">The connection.</param>
         /// <returns>The provider factory for the connection.</returns>
--- a/src/MiniProfiler.EF6/EFProfiledDbProviderFactoryResolver.cs
+++ b/src/MiniProfiler.EF6/EFProfiledDbProviderFactoryResolver.cs
@@ -1,0 +1,35 @@
+﻿using System.Data.Common;
+using System.Data.Entity.Infrastructure;
+using StackExchange.Profiling.Data;
+
+namespace StackExchange.Profiling.EntityFramework6
+{
+    /// <summary>
+    /// Wrapper for a service for obtaining the correct System.Data.Common.DbProviderFactory from
+    //  a given System.Data.Common.DbConnection.
+    /// </summary>
+    public class EFProfiledDbProviderFactoryResolver : IDbProviderFactoryResolver
+    {
+        private readonly IDbProviderFactoryResolver _inner;
+
+        /// <summary>
+        /// Initialises a new instance of the <see cref="EFProfiledDbProviderFactoryResolver"/> class.
+        /// </summary>
+        /// <param name="inner">The IDbProviderFactoryResolver to wrap.</param>
+        public EFProfiledDbProviderFactoryResolver(IDbProviderFactoryResolver inner)
+        {
+            _inner = inner;
+        }
+
+        /// <summary>
+        /// Returns the System.Data.Common.DbProviderFactory for the given connection,
+        /// unwrapping the ProfiledDbConnection as necessary
+        /// </summary>
+        /// <param name="connection">The connection.</param>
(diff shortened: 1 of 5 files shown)


src/MiniProfiler.EF6/MiniProfilerEF6.cs
```diff
@@ -10,15 +13,23 @@ namespace StackExchange.Profiling.EntityFramework6
     /// </summary>
     public static class MiniProfilerEF6
     {
+        private static readonly ConcurrentDictionary<object, object>  _resolvedDependenciesCache = new ConcurrentDictionary<object, object>();
```
NickCraver: Are we sure this doesn't grow unbounded?
I'd argue for 4 properly typed dictionaries here, assuming there's no unbounded growth. This also has a side-effect of caching the services even if they were to be replaced downstream later, is that a concern? e.g. the thing we're wrapping could be swapped, and we wouldn't care, as the `inner` isn't a key function.

src/MiniProfiler.EF6/MiniProfilerEF6.cs
```diff
@@ -35,5 +46,29 @@ public static void Initialize()
                 }
             }
         }
+
+        private static DbProviderServices WrapDbProviderServices(DbProviderServices inner, object key)
+        {
+            var cacheKey = new { type = typeof(DbProviderServices), key }; // TODO: consider using ValueTuple or implementing a similar polyfill
```
NickCraver: I'd use a ValueTuple here normally, but depending on the bounds I'd instead opt for 4 properly typed `ConcurrentDictionary` instead, which means only `key` is used anyway.

src/MiniProfiler.EF6/EFProfiledInvariantNameResolver.cs
```diff
@@ -0,0 +1,71 @@
...
+            var service = GetService(type, key);
+            return service == null ? Enumerable.Empty<object>() : new[] { service };
+        }
+
+        private static IProviderInvariantName GetProviderInvariantNameViaReflection(DbProviderFactory factory)
+        {
+            // Avert your eyes. EF6 implenents a handy helper method to get the Invariant Name given a DbProviderFactory instance,
+            // but of course it is marked internal. Rather than rewrite all of that code, we'll just call into it via reflection and cache the result.
+            var extensionsType = Type.GetType("System.Data.Entity.Utilities.DbProviderFactoryExtensions, EntityFramework");
+            var GetProviderInvariantNameMethod = extensionsType.GetMethod("GetProviderInvariantName", BindingFlags.Static | BindingFlags.Public);
```
NickCraver: These lines should be inside the `try` as well, no?

src/MiniProfiler.EF6/EFProfiledDbProviderFactoryResolver.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System.Data.Common;
+using System.Data.Entity.Infrastructure;
+using StackExchange.Profiling.Data;
+
+namespace StackExchange.Profiling.EntityFramework6
+{
+    /// <summary>
+    /// Wrapper for a service for obtaining the correct System.Data.Common.DbProviderFactory from
```
NickCraver: `<cref` on these

src/MiniProfiler.EF6/EFProfiledInvariantNameResolver.cs
```diff
@@ -0,0 +1,71 @@
...
+using StackExchange.Profiling.Data;
+
+namespace StackExchange.Profiling.EntityFramework6
+{
+    /// <summary>
+    /// Replacement for the DefaultInvariantNameResolver which can correctly resolve an IProviderInvariantName given a ProfiledDbProviderFactory.
+    /// </summary>
+    internal class EFProfiledInvariantNameResolver : IDbDependencyResolver
+    {
+        private static ConcurrentDictionary<DbProviderFactory, IProviderInvariantName> _providerInvariantNameCache = new ConcurrentDictionary<DbProviderFactory, IProviderInvariantName>();
```
NickCraver: Should this be static? Are we allocating a resolver per in this pattern?

src/MiniProfiler.EF6/MiniProfilerEF6.cs
jdaigle: > Are we sure this doesn't grow unbounded?

Fairly confident. In the case of `IDbProviderFactoryResolver` and `IDbConnectionFactory`, its treated as a singleton and the "key" is always null. For `DbProviderServices` and `DbProviderFactory`, the key is always the provider name.

> I'd argue for 4 properly typed dictionaries here

I like that, very clean. Wish I thought of it originally.

This style of caching actually mimics the caching that EF does (there's an internal `CachingDependencyResolver` that they use for the core resolvers of these same types which behave the exact same way).

This shouldn't change the behavior if something else replaces the service either before or after we do - however there's probably a good chance that if something else were replacing these services, things won't work well... I'm not sure we can totally solve that.

src/MiniProfiler.EF6/EFProfiledInvariantNameResolver.cs
jdaigle: Perhaps. `GetMethod()` won't throw `TargetInvocationException`, but `GetType()` *might*.

NickCraver:
Looking good now, thank you! Merged in :)

PR #191, opened 2017-06-20
Fixing a bug caused by attempting to cache EF dependencies with a null key

jdaigle:
EF may try to resolve dependencies with a null key for global singletons.

I'm annoyed at myself for not testing the refactoring of the original PR right way...


NickCraver:
Looks good, thanks!

PR #196, opened 2017-07-17
Ignore Duplicate based on ExecuteType (Fixes #194)

### LINKED ISSUE #194: `Connection Open()`/`Connection OpenAsync()` Marked as Duplicate
I don't necessarily think it's a problem when a single HTTP request happens to make multiple calls to  Open, and so this creates a lot of false-positives in the UI.

Since this is new behavior in v4, I think there are two ways to handle this:

1. Always ignore ExecuteType === "Open"/OpenAsync" when detecting duplicates in the UI.
2. A new setting (`IgnoreDuplicateOpenConnection`?) to conditionally control 1.

Whichever approach we like, I'll throw together a pull request.

jdaigle:
Introducing a new "CustomTimingExecuteTypesToIgnoreAsDuplicate" setting to
explicitly ignore duplicate custom timings based on its execute type.

By default it will ignore "Open" and "OpenAsync".

### DIFF
--- a/src/MiniProfiler.Shared/Internal/MiniProfilerExtensions.cs
+++ b/src/MiniProfiler.Shared/Internal/MiniProfilerExtensions.cs
@@ -98,6 +98,21 @@ namespace StackExchange.Profiling.Internal
             sb.Append("\" data-trivial-milliseconds=\"");
             sb.Append(MiniProfiler.Settings.TrivialDurationThresholdMilliseconds.ToString(CultureInfo.InvariantCulture));
 
+            if (MiniProfiler.Settings.IgnoredDuplicateExecuteTypes.Count > 0)
+            {
+                sb.Append("\" data-ignored-duplicate-execute-types=\"");
+                var i = 0;
+                foreach (var executeType in MiniProfiler.Settings.IgnoredDuplicateExecuteTypes)
+                {
+                    if (i > 0)
+                    {
+                        sb.Append(',');
+                    }
+                    sb.Append(executeType);
+                    i++;
+                }
+            }
+
             sb.Append("\"></script>");
 
             return sb.ToStringRecycle();
--- a/src/MiniProfiler.Shared/MiniProfiler.Settings.cs
+++ b/src/MiniProfiler.Shared/MiniProfiler.Settings.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Reflection;
+using StackExchange.Profiling.Data;
 using StackExchange.Profiling.Helpers;
 using StackExchange.Profiling.SqlFormatters;
 using StackExchange.Profiling.Storage;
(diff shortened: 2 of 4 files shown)


src/MiniProfiler.Shared/Internal/MiniProfilerExtensions.cs
```diff
@@ -34,7 +35,8 @@ public static class MiniProfilerExtensions
             bool? showTimeWithChildren = null,
             int? maxTracesToShow = null,
             bool? showControls = null,
-            bool? startHidden = null)
+            bool? startHidden = null,
+            HashSet<string> customTimingExecuteTypesToIgnoreAsDuplicate = null)
```
NickCraver: IMO accepting this here is overkill - I'd just make it a setting

src/MiniProfiler.Shared/MiniProfiler.Settings.cs
```diff
@@ -148,6 +149,15 @@ public static class Settings
             public static bool ShowControls { get; set; } = false;
 
             /// <summary>
+            /// Custom timing ExecuteTypes to ignore as duplicates in the UI.
+            /// </summary>
+            public static HashSet<string> CustomTimingExecuteTypesToIgnoreAsDuplicate { get; } = new HashSet<string>()
```
NickCraver: `CustomTimingExecuteTypesToIgnoreAsDuplicate` is a bit long, how about `IgnoredDuplicateExecuteTypes`?

src/MiniProfiler.Shared/MiniProfiler.Settings.cs
```diff
@@ -148,6 +149,15 @@ public static class Settings
...
+            {
+                nameof(DbConnection.Open),
```
NickCraver: Don't use `System.Data` here, instead use `nameof(ProfiledDbConnection.Open)` (what's actually used) - also makes "Find All References", etc. work :)

src/MiniProfiler.Shared/ui/includes.js
```diff
@@ -216,7 +216,7 @@ var MiniProfiler = (function () {
                     customTiming.ParentTimingId = timing.Id;
                     customStat.Duration += customTiming.DurationMilliseconds;
                     customStat.Count++;
-                    if (customTiming.CommandString && duplicates[customTiming.CommandString]) {
+                    if (customTiming.CommandString && duplicates[customTiming.CommandString] && !ignoreDuplicateCustomTiming(customTiming)) {
```
NickCraver: I'd change this to be simpler: remove this change and instead modify the `else` below to `else if (!ignoreDuplicateCustomTiming(customTiming))`, so it never gets added to `duplicates` below, that is less likely to get tripped on for uses of `duplicates` later.

jdaigle:
Implemented suggested changes and tested.

NickCraver:
Looks great, thanks @jdaigle!

PR #202, opened 2017-08-08
Added link to results-index page 

adamjez:
It would be nice to have a link to the **result-index** page in the miniprofiler "widget" so I do not have to remember the link or put it in my own site.

Result looks like this: 
newminiprofilerwidget

This PR also contains fix for vscode configurion (launch.json) and fixed link in AspNetCore sample project

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,9 +6,9 @@
             "type": "coreclr",
             "request": "launch",
             "preLaunchTask": "build",
-            "program": "${workspaceRoot}\\samples\\Samples.AspNetCore\\bin\\Debug\\netcoreapp1.0\\Samples.AspNetCore.dll",
+            "program": "${workspaceRoot}/samples/Samples.AspNetCore/bin/Debug/netcoreapp1.0/Samples.AspNetCore.dll",
             "args": [],
-            "cwd": "${workspaceRoot}",
+            "cwd": "${workspaceRoot}/samples/Samples.AspNetCore",
             "stopAtEntry": false,
             "internalConsoleOptions": "openOnSessionStart",
             "launchBrowser": {
--- a/samples/Samples.AspNetCore/Views/Home/Index.LeftPanel.cshtml
+++ b/samples/Samples.AspNetCore/Views/Home/Index.LeftPanel.cshtml
@@ -28,7 +28,7 @@
                     {
                     <li><a asp-controller="Test" asp-action="DisableProfilingUI">Disable profiling UI</a></li>
                     }
-                <li><a href="/mini-profiler-resources/results-index">Show all profiling sessions</a></li>
+                <li><a href="/profiler/results-index">Show all profiling sessions</a></li>
             </ul>
         </div>
     </div>


NickCraver:
I like the other fixes, but the custom link isn't something many people would want and importantly *is already customizable*, so users who want it can simply add it to their config.

Can you split out all the other fixes so that I can pull them in please?

adamjez:
I know there is a custom link, but I do not like adding the URL in each particular request. Or am I missing something? I did not find any global config

NickCraver:
@adamjez You can do so in your convenience method wrapping the `.Stop*()` methods and such, that's the typical behavior. The same way you don't like adding it, other people don't have *having* it (and many don't use that route at all), for example on Stack Overflow we wouldn't want that link at all, due to being a server farm isn't not a good indicator of anything and we have custom sampling for that. There would also be a disconnect in authentication on whether you can even see that page, so you're likely adding a 404 to many people as well.

adamjez:
@NickCraver Ok I get it. And What about adding the possibility to set also **global** custom link?

NickCraver:
@adamjez I'm not sure it's common enough to add a path for given this is the first request I've heard for it, so adding it to the case, IMO, is had to justify. You'd have to do one of:
1. Change the serializer to add the custom link on return to the client (not in each profiler's collection)
2. Actually add the link to every profiler's collection
3. Add a whole other path for this, which means wrapping the profiler and changing all client code (because remember what's sent to a client is just the profiler serialized, not another structure)

I think 3 is a pretty bad option for complexity right off. Option 2 eats storage unnecessarily, but is simple (against the tenants of a "Mini" profiler though). I'm not sure how number 1 would work. Presumably we'd have to override the serialization with a custom serializer when hitting the property to add the link, which may have downstream impact for anyone serializing the object directly as it'd behave differently (I'm not sure this is a large concern).

adamjez:
@NickCraver I most like the 2 option. I don't think the eaten storage is something that we should worry about. It would use the storage only if you add a global custom link. 

I'm really surprised that I'm the first who would like to have link to the history page somewhere. I believe a lot of people do not even know about this page. I did not find the results-index page anywhere in the **new** [documentation](http://miniprofiler.com/dotnet/AspDotNetCore), I learned about it through the source code.

NickCraver:
@adamjez I'll merge this on top of the de-static work in the other PR, not ignoring it. I thought about this more and here's what I think (bear with me):

This is a bad idea *as an extension point*. It's very specific and not all that useful for the vast majority case. So how about we aim at something easy to do this with? Something like an `OnBeforeSave` event handler on the options perhaps? This way people could do things like custom naming, add links, attach other profiler trees, or really whatever you wanted before saving.

For example your config case could be something like:
```c#
options.OnBeforeSave = profiler => profiler.AddCustomLink("Results", "<url>");
```
(that's a `Action<MiniProfiler>` example, but I'd play and see that vs. a multi-use event handler)

NickCraver:
Merging fixes in - how about we discuss the above proposal in a new issue?

PR #210, opened 2017-08-22
Options and de-static refactor

NickCraver:
Let's be honest: **this will break everyone**. But it's necessary to get where we need to be in simplicity. This will be a rough upgrade and needs some docs loving.

This is a series of changes that aims to:
- Decouple several concepts and interfaces from one another
- Make Options the root object a user needs, not many things
- Remove all statics except those needed for MiniProfiler.Current
- Be more friendly to initial setup (on both platforms)
- Ease option configurations
- Have more sensible defaults
- Simpify EntityFramework integration
- Increase testability (be removing most statics)
- Simplify ASP.NET Core configuration and follow the -isms there
- Make ASP.NET Core config *just work* with the defaults
- Bring parity to ASP.NET Core options (e.g. UserId default == IP)
- Simplify the `MemoryCacheStorage` bifurcation by moving downstream into the per-framework libraries (`IMemory` vs. non versions).

There's more to do here, this is just for the next alpha. Next up:
- .AspNetCore tests

This also upgrades xUnit to include my PR which allows the async tests to run at the end by themselves and be a lot more stable on timings for fewer-core machines.


src/MiniProfiler.AspNetCore.Mvc/MvcExtensions.cs
```diff
@@ -23,6 +25,17 @@ public static IMiniProfilerBuilder AddMiniProfiler(this IServiceCollection servi
...
         }
+
+        /// <summary>
+        /// Adds MiniProfiler timings for actions and views.
+        /// </summary>
+        /// <param name="services">The services collection to configure.</param>
+        /// <param name="configureOptions">An Action{MiniProfilerOptions} to configure options for MiniProfiler.</param>
+        public static IMiniProfilerBuilder AddMiniProfiler(this IServiceCollection services, Action<MiniProfilerOptions> configureOptions = null)
+        {
+            services.Configure(configureOptions);
```
nickalbrecht: The ASPNET Classic MiniProfiler doesn't appear to depend on any explicitly defined configuration from the web.config. If the final ASPNET Core implementation is planning to do the same, and it doesn't depend on anything defined in a configuration (such as a json file). There's really no reason to use `Configure()` here. In fact you might be able to eliminate a direct dependency on `Microsoft.Extensions.Options` by using `TryAddSingleton()` or the like instead. Using Configure extension method just allows you to benefit from things offered by the Configuration like refreshing on changes, binding against a source, and being able to composite the resulting effective configuration from multiple disparate files & formats.

src/MiniProfiler.AspNetCore/MiniProfilerMiddleware.cs
```diff
@@ -36,11 +36,11 @@ public class MiniProfilerMiddleware
         public MiniProfilerMiddleware(
             RequestDelegate next,
             IHostingEnvironment hostingEnvironment,
-            MiniProfilerOptions options)
+            IOptions<MiniProfilerOptions> options)
         {
             _next = next ?? throw new ArgumentNullException(nameof(next));
             _env = hostingEnvironment ?? throw new ArgumentNullException(nameof(hostingEnvironment));
-            Options = options ?? throw new ArgumentNullException(nameof(options));
+            Options = options.Value ?? throw new ArgumentNullException(nameof(options));
```
nickalbrecht: I commented on this elsewhere about potentially changing the registration of the options to not use `Configure()` and `IOptions<>` if none of the options actually depend on pulling information from a configuration source. But it's worth pointing out that should you not decide to do this and instead keep `Configure`/`IOptions`, that you should not save the result of the `Value` property beyond it's use. This is what allows your options to be refreshed should the source configuration be notified of a change. By saving the value, you can run into a situation where your internal Options property in this class will be what it was when the app first started, but requesting `IOptions<MiniProfilerOptions>` again from the ServiceProvider yields an options instance with Value being a different instance entirely.

src/MiniProfiler.AspNetCore.Mvc/MvcExtensions.cs
NickCraver: I've just pushed latest which hopefully means the user very rarely has to do either, this was a mess (to the user), IMO. Hopefully latest is much cleaner all around. It's using DI on the configure for the IMemoryCache, which also means the user need not have a separate `.AddMemoryCache()` if going with app-local-in-memory defaults.

src/MiniProfiler.AspNetCore/MiniProfilerMiddleware.cs
NickCraver: Thanks for the tips here, I think I addressed both in latest.

We *may* have something loading from JSON earlier (e.g. a storage configed location), so that's the reason for `.Configure()` now (in relation to your other comment).

PR #216, opened 2017-09-14
Corrects the path to the Mvc5 Sample

Romanx:


NickCraver:
Doh, thanks for the fix!

PR #221, opened 2017-09-30
NetCore: Add a new SQLite Provider

ananth-racherla:
This PR introduces a new SQlite provider to be use for .NET Core projects.  It is structured very similar to the existing SQLite provider, except it uses Entity Framework.

However the primary motivation was to fix a "Null Reference" error which occurs within RenderIncludes function for Asp.net Core projects.  

When Storage.Load is called, the MiniProfiler.Options attribute is not "rehydrated" by any of the existing storage providers (except for memory storage provider).  This causes the null exception in RenderIncludes as mentioned.

The workaround used in this PR  was to create a new MiniProfiler with default options when Storage.Load is called.   I am not sure if there is a better way to deal with this.  

Rather than logging a bug, I hoping this PR might be better able to highlight the issue

### DIFF
--- a/src/MiniProfiler.Providers.Sqlite.EF/SqliteMiniProfilerStorage.cs
+++ b/src/MiniProfiler.Providers.Sqlite.EF/SqliteMiniProfilerStorage.cs
@@ -1,0 +1,320 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using StackExchange.Profiling.Internal;
+using System.Data.Common;
+using Microsoft.EntityFrameworkCore;
+
+namespace StackExchange.Profiling.Storage
+{
+    // <summary>
+    /// The SQLITE mini profiler storage.
+    /// </summary>
+    public class SqliteMiniProfilerStorage : DatabaseStorageBase
+    {
+        public IDataContextFactory ContextFactory { get; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SqliteMiniProfilerStorage"/> class.
+        /// </summary>
+        /// <param name="connectionString">The connection string.</param>
+        public SqliteMiniProfilerStorage(string connectionString, bool recreate = false) : base(connectionString)
+        {
+            ContextFactory = new ProfilerDbContextFactory(connectionString);
+
+            using (var dbContext = GetDbContext())
+            {
+                dbContext.Init(recreate);
+            }
+        }
+
+        private ProfilerDbContext GetDbContext()
+        {
+            return ContextFactory.Create();
+        }
+
+        /// The list of results.
+        /// </summary>
+        /// <param name="maxResults">The max results.</param>
+        /// <param name="start">The start</param>
+        /// <param name="finish">The finish</param>
+        /// <param name="orderBy">The order by.</param>
+        /// <returns>The result set</returns>
+        public override IEnumerable<Guid> List(int maxResults, DateTime? start = null, DateTime? finish = null, ListResultsOrder orderBy = ListResultsOrder.Descending)
+        {
+            var task = ListAsync(maxResults, start, finish, orderBy);
+            task.Wait();
+            return task.Result;
+        }
+
+        public async override Task<IEnumerable<Guid>> ListAsync(int maxResults, DateTime? start = null, DateTime? finish = null, ListResultsOrder orderBy = ListResultsOrder.Descending)
+        {
+            using (var DbContext = GetDbContext())
+            {
+                var query = DbContext.MiniProfilers
+                    .Where(w => (finish == null || w.Started < finish.Value) && (start == null || w.Started > start.Value));
+
+                query = orderBy == ListResultsOrder.Ascending ? query.OrderBy(o => o.Started) : query.OrderByDescending(o => o.Started);
+
+                return await query.Take(maxResults).Select(s => s.Id).ToListAsync();
+            }
+        }
+
+        /// <summary>
+        /// Stores to <c>dbo.MiniProfilers</c> under its <see cref="MiniProfiler.Id"/>;
+        /// </summary>
+        /// <param name="profiler">The <see cref="MiniProfiler"/> to save.</param>
+        public override void Save(MiniProfiler profiler)
+        {
+            SaveAsync(profiler).Wait();
+        }
+
+        /// <summary>
+        /// Asynchronously stores to <c>dbo.MiniProfilers</c> under its <see cref="MiniProfiler.Id"/>.
+        /// </summary>
+        /// <param name="profiler">The <see cref="MiniProfiler"/> to save.</param>
+        public async override Task SaveAsync(MiniProfiler profiler)
+        {
+            using (var DbContext = GetDbContext())
+            {
+                if (!DbContext.MiniProfilers.Any(a => a.Id == profiler.Id))
+                {
+                    await DbContext.MiniProfilers.AddAsync(GetMiniProfilers(profiler));
+                }
+
+                var timings = new List<Timing>();
+                if (profiler.Root != null)
+                {
+                    profiler.Root.MiniProfilerId = profiler.Id;
+                    FlattenTimings(profiler.Root, timings);
+                }
+
+                if (timings.Any())
+                {
+                    var newtimings = timings.Where(a => !DbContext.MiniProfilerTimings.Any(b => b.Id == a.Id)).ToList();
+                    await DbContext.MiniProfilerTimings.AddRangeAsync(newtimings.Select(timing => GetMiniProfilerTimings(timing)));
+                }
+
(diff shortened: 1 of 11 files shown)


src/MiniProfiler.Providers.Sqlite.EF/SqliteMiniProfilerStorage.cs
```diff
@@ -0,0 +1,320 @@
...
+        public override MiniProfiler Load(Guid id)
+        {
+            var task = LoadAsync(id);
+            task.Wait();
+            return task.Result;
+        }
+
+        private MiniProfiler MapOptions(MiniProfiler miniProfiler)
+        {
+            return miniProfiler == null ? null : new MiniProfiler(miniProfiler.Name, new MiniProfilerBaseOptions())
```
ananth-racherla: This is the workaround for null reference within  RenderInclues

NickCraver:
I'm just checking the intent here - is the motivation to highlight the null ref issue or to add a new provider? I'll dig into the null ref issue, appreciate the heads up on that :)

ananth-racherla:
Both :)   Apologies, I haven't paid attention to see if there are any contributor guidelines.  Please feel free to point to them.

NickCraver:
@ananth-racherla On the bug: gotcha. And thanks! I agree this needs to change and I'll take at how to restructure for next alpha.

On the new provider I guess I have to ask (sorry if this seems blunt, not my intent): why? Can you share thinking on the advantage of having an Entity Framework Sqlite provider when we already have one that works also based on connection string? In my view: the user just wants a library that connects to SQLite and handles profilers, given we already have a more flexible (e.g. configurable table names, creation scripts, etc.) and faster (Dapper over EF) version in play, I'm not sure there's any user wins to the selection/confusion here...unless I'm missing something?

ananth-racherla:
This really started as an exercise to analyze the problems I was facing trying to integrate this with asp.net core application. I was running into some compatibility issues (.netstandard 1.6). That may very well have been due to my environment, I am not sure.  I agree that there is no clear advantage for the current approach except for readability. I also agree with you that existing solution probably has better performance . I have not benchmarked this approach.

One other slight modification I did was to allow the Sqlite database to be cleared on Start, which was a particular use case I wanted. Not sure if this makes sense for a wider audience.

Not tied to this PR in anyway.  Happy to have just the null reference issue fixed.

NickCraver:
@ananth-racherla Gotcha, thanks! I have the create/drop schemas exposed and wired up in the test project now but I've debated doing it in another way so that the providers expose the methods. I've gone back and forth on this - if you think it'd be useful I have some ideas on how to expose it (without making .Shared) take any more dependencies.

ananth-racherla:
@NickCraver do you want me to close out this PR?

NickCraver:
@ananth-racherla Let's get this bug fixed first :) I'm trying to understand when you're hitting it - is it a `.RenderIncludes()` call *not* on `MiniProfiler.Current` (which should be in-memory)?

ananth-racherla:
Yes that is correct.  And for Asp.net core app at least that happens if you use any Storage other than In Memory Store.   And this happens because the "Options" attribute is null after the Storage.Load() call happens in such cases.
Try to pull up a specific result (/results?id=<guid>) and you will see it.

NickCraver:
@ananth-racherla A fixed release available on MyGet shortly should resolve the errors you saw, can you please give it a try and let me know?

The version with the fix is `4.0.0-alpha8-193`. This is called automatically in the ASP.NET Core code, but requires an explicit `MiniProfiler.Configure()` call in the older framework (samples are all updated). I'll ponder on making this cleaner but wanted to address your issue as soon as I could.

ananth-racherla:
Hello @NickCraver that seems to have fixed the null reference issue.  However the SQLiteStorage is throwing a cast exception:
```
System.InvalidCastException: Object must implement IConvertible.
   at System.Convert.ChangeType(Object value, Type conversionType, IFormatProvider provider)
   at Dapper.SqlMapper.<QueryAsync>d__23`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at StackExchange.Profiling.Storage.SqliteStorage.<ListAsync>d__20.MoveNext() in C:\projects\dotnet\src\MiniProfiler.Providers.Sqlite\SqliteStorage.cs:line 343
```
I suspect it doesn't quite like the Guid field somehow.  I did not see this issue though with my sqlite provider

ananth-racherla:
@NickCraver tested today with the Redis provider instead.  Seems to work fine with that.

NickCraver:
@ananth-racherla Finally found time today to dig into everything here and get tests in order - the latest MyGet feed should have a working Sqlite implementation if you have time to give it a go :)

NickCraver:
Closing this out to cleanup, please ping me if you have issues with the provider!

ananth-racherla:
@NickCraver apologies I could only get around to testing this today.   I pulled the latest versions of from 
MiniProfiler.AspNetCore.Mvc & MiniProfiler.Providers.Sqlite (4.0.0-alpha9-00215) from MyGet.

I am still seeing issues with the Sqlite profiler.  It seems to store the results correctly, but fails to read them back from the sqlite store.  This is the exception I see when I load up the /results-index endpoint.
```
System.InvalidOperationException: Error parsing column 5 (StartMilliseconds=54.7 - Double) ---> System.InvalidCastException: Unable to cast object of type 'System.Double' to type 'System.Int64'.
   at Deserialized159d169-aa79-4c19-b4fd-1f4e34d3b2db(IDataReader )
   --- End of inner exception stack trace ---
   at Dapper.SqlMapper.ThrowDataException(Exception ex, Int32 index, IDataReader reader, Object value)
   at Deserialized159d169-aa79-4c19-b4fd-1f4e34d3b2db(IDataReader )
   at Dapper.SqlMapper.GridReader.<ReadDeferred>d__52`1.MoveNext()
   at System.Collections.Generic.List`1.AddEnumerable(IEnumerable`1 enumerable)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Dapper.SqlMapper.GridReader.ReadImpl[T](Type type, Boolean buffered)
   at StackExchange.Profiling.Storage.SqliteStorage.Load(Guid id)
   at StackExchange.Profiling.MiniProfilerMiddleware.<ResultsListAsync>b__15_1(Guid g)
   at System.Linq.Enumerable.SelectEnumerableIterator`2.MoveNext()
   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeList(JsonWriter writer, IEnumerable values, JsonArrayContract contract, JsonProperty member, JsonContainerContract collectionContract, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.Serialize(JsonWriter jsonWriter, Object value, Type objectType)
   at Newtonsoft.Json.JsonSerializer.SerializeInternal(JsonWriter jsonWriter, Object value, Type objectType)
   at Newtonsoft.Json.JsonConvert.SerializeObjectInternal(Object value, Type type, JsonSerializer jsonSerializer)
   at StackExchange.Profiling.MiniProfilerMiddleware.<ResultsListAsync>d__15.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at StackExchange.Profiling.MiniProfilerMiddleware.<HandleRequest>d__11.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at StackExchange.Profiling.MiniProfilerMiddleware.<Invoke>d__7.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Hosting.Internal.RequestServicesContainerMiddleware.<Invoke>d__3.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.Frame`1.<ProcessRequestsAsync>d__2.MoveNext()
```

PR #236, opened 2017-10-31
Switch (back) to MySqlConnector

bgrainger:
This enables async database operations. (MySqlConnector has a fully async implementation; MySql.Data's async methods run synchronously on the caller's thread then return a completed Task.)

v0.30.0 is now strong-named, fixing the open issue with #188.


NickCraver:
@bgrainger awesome to have this back in the fold, thanks!

PR #237, opened 2017-11-05
Avoid hard-coded path to roslyn compiler

dustinsoftware:
This fixes the build on a fresh install of VS2017 Community


NickCraver:
Ah neat and what I was thinking of but couldn't actually do in the first place, thank you!

PR #244, opened 2017-12-11
Expose disposed inner connection

### LINKED ISSUE #243: EF6.2 "Value cannot be null. Parameter name: connection"
I'm using **MiniProfiler.EF6** 4.0.0-alpha9-00215, which is the latest version available on MyGet right now.

After calling `MiniProfilerEF6.Initialize()`, the following exception is thrown when querying the database:

```
Value cannot be null. Parameter name: connection
   at System.Data.Entity.Utilities.Check.NotNull[T](T value, String parameterName)
   at System.Data.Entity.Infrastructure.DefaultDbProviderFactoryResolver.ResolveProviderFactory(DbConnection connection)
   at StackExchange.Profiling.EntityFramework6.EFProfiledDbProviderFactoryResolver.ResolveProviderFactory(DbConnection connection) in C:\projects\dotnet\src\MiniProfiler.EF6\EFProfiledDbProviderFactoryResolver.cs:line 28
   at System.Data.Entity.Utilities.DbConnectionExtensions.GetProviderFactory(DbConnection connection)
   at System.Data.Entity.Core.Common.DbProviderServices.GetProviderFactory(DbConnection connection)
   at System.Data.Entity.Migrations.DbMigrator.CreateConnection()
   at System.Data.Entity.Migrations.DbMigrator.EnsureDatabaseExists(Action mustSucceedToKeepDatabase)
   at System.Data.Entity.Migrations.DbMigrator.Update(String targetMigration)
   at System.Data.Entity.MigrateDatabaseToLatestVersion`2.InitializeDatabase(TContext context)
   at System.Data.Entity.Internal.InternalContext.<>c__DisplayClassf`1.<CreateInitializationAction>b__e()
   at System.Data.Entity.Internal.InternalContext.PerformInitializationAction(Action action)
   at System.Data.Entity.Internal.InternalContext.PerformDatabaseInitialization()
   at System.Data.Entity.Internal.LazyInternalContext.<InitializeDatabase>b__4(InternalContext c)
   at System.Data.Entity.Internal.RetryAction`1.PerformAction(TInput input)
   at System.Data.Entity.Internal.LazyInternalContext.InitializeDatabaseAction(Action`1 action)
   at System.Data.Entity.Internal.LazyInternalContext.InitializeDatabase()
   at System.Data.Entity.Internal.InternalContext.Initialize()
   at System.Data.Entity.Internal.InternalContext.GetEntitySetAndBaseTypeForType(Type entityType)
   at System.Data.Entity.Internal.Linq.InternalSet`1.Initialize()
   at System.Data.Entity.Internal.Linq.InternalSet`1.AsNoTracking()
   at System.Data.Entity.Infrastructure.DbQuery`1.AsNoTracking()
```

I can only guess MiniProfiler wraps the database connection with a profiling layer which appears to contain a bug.

This worked fine in 4.0.0-alpha6-83. I am using .NET Framework 4.7.

rlenders:
Fixes #243. At some point IDbProviderFactoryResolver is asked to resolve a factory for a disposed profiled connection. The profiled implementation of this interface unwraps the profiled connection and passes the inner connection to the actual factory resolver. This caused "Value cannot be null. Parameter name: connection" because the inner connection was null after disposing. I don't see why the inner connection should be set to null after disposing, as it can still be useful. Such as for resolving factories based on connections.

### DIFF
--- a/src/MiniProfiler.Shared/Data/ProfiledDbConnection.cs
+++ b/src/MiniProfiler.Shared/Data/ProfiledDbConnection.cs
@@ -160,7 +160,6 @@ namespace StackExchange.Profiling.Data
                 _connection.Dispose();
             }
             base.Dispose(disposing);
-            _connection = null;
             _profiler = null;
         }
 


NickCraver:
I need to dig into your error and what's throwing - I don't agree with the fix here as the entire point of disposing is to release resources. The longer a connection is held here, the longer it's kept from being returned to the pool, etc. I am curious why you're the first one (to my knowledge) that's hit such an error, there may be something different about your overall setup that needs to be factored in.

rlenders:
I figured it out. I'll close this PR and elaborate on my findings in #243.

Issue #252, opened 2018-01-08
StackExchange.Profiling.Wcf outdated


Danielku15:
I'd like to ask what happened to the StackExchange.Profiling.Wcf package. It appears to be quite outdated compared to the other packages. 
  
Also it appears that a new version of MiniProfiler is in active development, are there plans to still support WCF? I couldn't find any source on WCF on the master branch. I would have a small addition/fix to WCF on v3: Support of duplex channels and proper action names in the profiler name. I was wondering if there is a change for an official nuget package if I contribute those changes.

NickCraver:
I don't (currently) have any plans to support WCF (or familiarity with it...that's the reason). But, if someone wanted to PR an update to 4.0 that adds WCF support in the new version I'd be more than happy to take a look and add it in :)

Danielku15:
Is there already a plan when 4.0 will be released? I'd be happy to add the functionality so we can use then also 4.0 in our product. I think 90% of the code can anyhow be reused :smile:

NickCraver:
@Danielku15 I'm back to working on 4.0 now (had to take a break to build a basement), hopefully going out over the next few weeks as RTM. I'm looking into open issues now. WCF will follow the 4.0 release if contributed afterwards, but happy to take a look at a PR w/ sample project additions at any point - it'd be much appreciated.

NickCraver:
@Danielku15 The beta for 4.0 is now on NuGet if this is still appealing to port!

NickCraver:
Going to close this out since there's been no other activity or interest here. My current assumption is people on WCF are using MiniProfiler v3.2, or are capable of sticking to that version.

PR #253, opened 2018-01-09
Added support for callback channels and improved profiler naming

Danielku15:
This PR adds support for WCF callback channels by 

- Registering both Client and Dispatch inspectors on server and client side
- Improved handling of profiler name generation 

Related to #252

### DIFF
--- a/StackExchange.Profiling.Wcf/WcfMiniProfilerBehavior.cs
+++ b/StackExchange.Profiling.Wcf/WcfMiniProfilerBehavior.cs
@@ -33,8 +33,12 @@
         /// <param name="clientRuntime">The client runtime.</param>
         public void ApplyClientBehavior(ServiceEndpoint endpoint, System.ServiceModel.Dispatcher.ClientRuntime clientRuntime)
         {
-            var inspector = new WcfMiniProfilerClientInspector();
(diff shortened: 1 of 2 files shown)


NickCraver:
I'm torn here because it's a ton of work to do an update on the v3 branch of code for all. But at the same time, I'm not sure how much work it is to a) bring the WCF library up to v4 compat or b) how much WCF is used ("is `a` worth it?").

Can you give me some idea of the platforms and versions you're using MiniProfiler on? I've gotten minimal feedback on dropping WCF in v4. Admittedly, I'm no expert on WCF, not even remotely familiar with it. So, I'm also one of the worst people to maintain such a thing. I'd welcome the addition to v4 but I'm not sure what the level of interest it and if anyone is open to contributing there. For example I could possibly get it up to date, but it also be buggy as hell and incredibly inefficient from lack of familiarity on my end.

Any info you can provide here would be helpful. I'd like to support as many people as possible with the resources I have, but I can't see that including supporting the v3 branch :( If it's worth the time (and maybe it's not that much time), then getting WCF support in v4 is the way to do. Thoughts?

NickCraver:
No news on this - so going to close out to cleanup.

PR #257, opened 2018-01-19
Fix #256 - Parameter name or value could break regex

### LINKED ISSUE #256: InlineFormatter throws when parameter contains parameter-like string
EDIT: v4

Repro:
```
var formatter = new StackExchange.Profiling.SqlFormatters.InlineFormatter();

var parameters = new List<SqlTimingParameter>();
	
parameters.Add(new SqlTimingParameter() { DbType = "string", Name = "url", Value = "http://www.example.com?myid=1" });
parameters.Add(new SqlTimingParameter() { DbType = "string", Name = "myid", Value = "1" });

var command = @"SELECT * FROM urls WHERE url = @url OR myid = @myid";

formatter.FormatSql(command, parameters);
```

If I change the ordering of the parameter it works.

The problem is that the ParamPrefix-regex matches "?myid" from the url (that has already been inserted into commandText).

That, in turn, causes an exception on line 47 because of an invalid grouping.
A Regex.Escape around "name" on line 47 fixes the exception,
but the output will then be: 
"SELECT * FROM urls WHERE url = 'http://www.example.com1=1' OR myid = @myid"

Adding the Regex.Escape and keeping a copy of the original commandtext (for finding the parameter prefix) solves the issue.

```
var originalCommandText = commandText;

foreach (var p in parameters)
            {
                // If the parameter doesn't have a prefix (@,:,etc), append one
                var name = ParamPrefixes.IsMatch(p.Name)
                    ? p.Name
                    : Regex.Match(originalCommandText, "([@:?])" + p.Name, RegexOptions.IgnoreCase).Value;
                if (name.HasValue())
                {
                    var value = GetParameterValue(p);
                    commandText = Regex.Replace(commandText, "(" + Regex.Escape(name) + ")([^0-9A-z]|$)", m => value + m.Groups[2], RegexOptions.IgnoreCase);
                }
            }
```

Don't you just love a weird regex edge case? ;)

I'll see if I can make a PR tomorrow

andersjonsson:
A parameter name or value that contained regex structures would make the regex break.
A part of a parameter value could also be identified as the name of a following parameter.

Fix #256

### DIFF
--- a/src/MiniProfiler.Shared/SqlFormatters/InlineFormatter.cs
+++ b/src/MiniProfiler.Shared/SqlFormatters/InlineFormatter.cs
@@ -35,16 +35,18 @@ namespace StackExchange.Profiling.SqlFormatters
                 return commandText;
             }
 
+            var originalCommandText = commandText;
+
             foreach (var p in parameters)
             {
                 // If the parameter doesn't have a prefix (@,:,etc), append one
                 var name = ParamPrefixes.IsMatch(p.Name)
                     ? p.Name
-                    : Regex.Match(commandText, "([@:?])" + p.Name, RegexOptions.IgnoreCase).Value;
+                    : Regex.Match(originalCommandText, "([@:?])" + Regex.Escape(p.Name), RegexOptions.IgnoreCase).Value;
                 if (name.HasValue())
                 {
                     var value = GetParameterValue(p);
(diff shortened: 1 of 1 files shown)


NickCraver:
Change looks good - I'm adding a unit test as well to ensure no regressions, thanks!

NickCraver:
Test added in 8cf7e547c42262d9b88c722f63b1909d1eddeff1 - thanks again for filing and a fix to boot!

PR #261, opened 2018-02-07
Update includes.js to prefer standardized timing APIs instead of deprecated ones

karl-sjogren:
Includes.js is using the window.chrome.loadTimes() API which has been marked as deprecated in the latest Chrome version and warns about it in the console.

``[Deprecation] chrome.loadTimes() is deprecated, instead use standardized API: Paint Timing.``

This PR first checks if the new API is available (it is from Chrome 60) or it will fall back to the old deprecated one.

The new API doesn't support ``firstPaintAfterLoadTime`` though but it seems that like [Chrome never did](https://developers.google.com/web/updates/2017/12/chrome-loadtimes-deprecated#firstpaintafterloadtime) anyway. It does however add "first-contentful-paint" which is a way more interesting value to keep track of.

### DIFF
--- a/src/MiniProfiler.Shared/ui/includes.js
+++ b/src/MiniProfiler.Shared/ui/includes.js
@@ -106,14 +106,27 @@ var MiniProfiler = (function () {
                     }
                     clientPerformance = copy;
 
-                    // hack to add chrome timings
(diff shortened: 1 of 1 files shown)


NickCraver:
A good fix indeed, thank you!

PR #264, opened 2018-02-26
Fix firstPaintAfterLoadTime check

dchaib:
I think there was a small typo for "First Paint After Load Time" timing.


NickCraver:
Good catch, thanks!

PR #266, opened 2018-02-27
`i` is already used in the function

andyedinborough:
This can cause an endless loop, because `i` is overwritten.


NickCraver:
Thanks so much for tracking this down, I'm very limited on time with other fires and this was a huge time saver <3

andyedinborough:
No problem.  Glad I could help!  ;]

PR #268, opened 2018-03-18
Simplify mongo db provider

DaniilSokolyuk:
Hi, @NickCraver, i saw your tweet and I have some ideas
1. Only 1 colleciton
2. No poco
3. Manual mapping
4. Fix bug https://github.com/MiniProfiler/dotnet/blob/c32b23097331666637b846d6b83389363be29cd5/src/MiniProfiler.Providers.MongoDB/MongoDbStorage.cs#L94 to Less Than 
5. Add WithIndexCreation method

### DIFF
--- a/src/MiniProfiler.Providers.MongoDB/MongoDbStorage.cs
+++ b/src/MiniProfiler.Providers.MongoDB/MongoDbStorage.cs
@@ -82,6 +82,13 @@ namespace StackExchange.Profiling
                 });
         }
 
+        public MongoDbStorage WithIndexCreation()
+        {
+            _collection.Indexes.CreateOne(Builders<MiniProfiler>.IndexKeys.Ascending(_ => _.Started));
+            _collection.Indexes.CreateOne(Builders<MiniProfiler>.IndexKeys.Ascending(_ => _.HasUserViewed));
+            return this;
+        }
+
         /// <summary>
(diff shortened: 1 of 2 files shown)


NickCraver:
Well this is just a beautiful thing to wake up to! I thought Mongo could be much simpler but haven't looked in ages. I very much appreciate the drastic simplification here, it's far less to maintain and obviously far easier to work with. Do you mind a few questions? I'm wondering on:

- Should we implement IDiposable at all? At a glance I didn't see any disposal code in the client or database...do I miss other things Mongo traditionally wants disposed in the driver?
- What happens if/when we add a field? Is adding it to the mapping all that's needed to play nice with additions? *If* that's not an issue, I'd *assume*: non-nullables are defaults and nullables are null...am I off there?

Thanks! I'm happy to merge this in when you're happy with the state, and obviously I should go add a lot more detail to the tests to all storage providers...the initial merge passed checks with no timings but now that we have the provider shape I can do make those very detailed with no runtime cost and make sure we're all good with any changes across the board of providers.

DaniilSokolyuk:
- Yes, we can remove IDispasable
- You right, non-nullables are defualts and nullables are null, but if you change field type you will fall with an error on deserialization
I added some indexes, I think you can merge

PR #276, opened 2018-05-06
Client cleanup: TypeScript/no-.tmpl conversion

### LINKED ISSUE #235: Content security policy - unsafe-eval
Hey,

I'm using MiniProfiler in an application with a pretty restrictive `Content-Security-Policy`. At present, the `script-src` directive does not permit the `unsafe-eval` source which breaks the MiniProfiler UI.

Would it be possible to change the templates to avoid this sort of code?

```
Error: call to Function() blocked by CSP    includes.js:965:4245
Content Security Policy: The page’s settings blocked the loading of a resource at self (“default-src https://app.local:443”). Source: call to eval() or related function blocked by CSP.     includes.js:965
```

NickCraver:
There's a lot going on here, but mostly a massive client-side cleanup in general. Of note:
- Moves the lib to TypeScript with a bundler for creating `includes.js` with `jQuery`, `jQuery.hotkeys`, and `highlight.js` included.
  - This bundle happens in the build
  - `includes.js` is zeroed out (since we build it now)
  - tslint is configured
  - ES5 is the current target, for a reasonable browser support matrix
  - `npm` is used for definitions/`@types`
  - Lots of cleanup of the general code (repeated loops, etc.) but still more to go
- No more `jQuery.tmpl()` plugin usage
  - This means no more `eval()` as well (fixes #235)
  - Heavy usage of TypeScript/ES6 multi-line literals and format strings. Not awesome, but better than the template plugin
  - No more usage of `localStorage`
- Highlighting has moved from `prettyPrint()` to `highlight.js` (but it's pluggable, via `MiniProfiler.highlight = ...`)
- CSS classes have moved from `.profiler-` to `.mp-`, to be shorter and have less conflict.
- JSON parses the `.Started` date to a date as it comes in

Semi-related bugs fixed during this:
- .NET SDK is pegged at latest stable, since preview builds are broken in all sorts of fun ways
- Height of the profiler popup while scrolled is now fixed

Note: this was a TypeScript learning experience. It's probably all wrong in terms of module layout (would love for someone to give feedback here), but everything's working and tested as far as I can tell.
Note 2: while this looks like a large line count increase, that's only in definition files, client-side is roughly what it was before...just easier to maintain now.

I think getting off jQuery isn't insane at this point either...need to explore that a bit more.


NickCraver:
Added visualization to client timings to make it a bit faster to grep:
image

PR #278, opened 2018-06-02
Use HTML escape instead of javascript escape function

adamjez:
Currently, If I use URL as profiler name I get to see eg. `http%3A//localhost%3A53168/` in `results-index` page.

I have replaced javascript [escape](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/escape) function, which is is unnecessarily strict, with HTML escape function.


NickCraver:
This was a rename artifact, definitely not intentional. Thanks for the fix!

PR #282, opened 2018-06-04
Fixing JavaScript issue when StackTraceSnippet property is undefined

jdaigle:
When `options.ExcludeStackTraceSnippetFromCustomTimings = true;`, the `StackTraceSnippet` of the custom timing is null. The null value is omitted from the emitted JSON.

Results in `Uncaught TypeError: Cannot read property 'replace' of undefined` when attempting to HTML encode the property.

### DIFF
--- a/src/MiniProfiler.Shared/ui/lib/MiniProfiler.ts
+++ b/src/MiniProfiler.Shared/ui/lib/MiniProfiler.ts
@@ -812,7 +812,7 @@ namespace StackExchange.Profiling {
               </td>
               <td>
                 <div class="query">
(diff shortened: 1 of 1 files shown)


NickCraver:
It seems like we should more generally handle this in `encode`, don't ya think?

NickCraver:
@jdaigle Appreciate the report - just moved the fix a bit in 1c0fe4486be63ff1a6e67301c2026683f33476f9 - handling the general case rather than just the one...because I bet that bites us again :)

jdaigle:
@NickCraver thanks! Admittedly I just went for the fastest/smallest fix. Your fix is better. :)

PR #283, opened 2018-06-04
Display "show trivials" when any client timing is trivial

adamjez:
Currently, when all _server_ timings are not trivial, the `show trivials` link is hidden even if there is trivial client timings


NickCraver:
Good fix, thanks!

PR #284, opened 2018-06-04
Added custom init condition

adamjez:
I would like to be able to override init condition. Sometimes the *PopupView* is displayed earlier than all my client timings are finished.

eg.: 
image
In this example, I'd like to show *PopupView* after DotVVM is initialized. Sometimes it displays the *PopupView* before DotVVM initialization end. 

I'm not sure If I do something wrong or If I use it the way it was not intended to.


NickCraver:
Not against this at all - do you have an example usage, e.g. what it'd look like in the page? If we could one-liner that, leaving it as a comment beside the option would be good.

NickCraver:
@adamjez I think that's just fine honestly - let's add a `// window.MiniProfiler.initCondition = function() { return myOtherThingIsReady; };` as an example comment after :)

NickCraver:
Added and merged, thanks!

PR #285, opened 2018-06-05
Fix showControls when jQuery is not included

Diego-T:
If you don't already include jQuery in your page and use the `show-controls="true"` option you get the following error:
```
Uncaught TypeError: $ is not a function
    at n.initControls (includes.min.js?v=4.0.0+g9bca957cdc:2)
....
```


NickCraver:
Dammit, thought I had gotten them all but added jQuery back to sample projects too early :( Thanks for the fix!

Issue #184, opened 2017-05-09
Can't run Samples.Mvc5 because it's a class library


RudeySH:
Title says it all, Samples.Mvc5 is a class library, it does not have a Program.cs with a static Main entry point. How do I run this sample?

jdaigle:
Technically it's an ASP.NET application (i.e. ASP.NET 4.x, which requires IIS). But the project is using the new .NET SDK project schema which doesn't correctly support launching IIS or IISExpress for ASP.NET 4.x applications from within Visual Studio.

The workaround is to manually create an IIS site and point it at the source directory.

NickCraver:
Yep - nothing we can do here, unfortunately. I hope the .NET guys will add support back in VS 2017 soon.

cc @davkean - any idea if there's a timeline for MVC5 support in new `.csproj`, or if it's been cut?

NickCraver:
I have opened up https://github.com/dotnet/project-system/issues/2670 in hopes they'll add the functionality to handle all cases like this.

Daniel15:
The way I handle this in one of my projects (ReactJS.NET) is that the MVC 5 sample uses the "old style" .csproj, while every other project uses the new-style ones. It works fine.

NickCraver:
While not a long-term solution (they need to just add support) a fix for the starting specifically can be found here: https://github.com/aspnet/Home/issues/1978#issuecomment-330894604

PR #288, opened 2018-06-07
Use a class to toggle trivial gaps

xt0rted:
The `.toggle()` function sets `display: block` so the table row doesn't show correctly. This fixes that. I also nuked a couple tabs for good measure.

There's already a `.show-trivial` class so I didn't want to reuse that name unless you wanted to.

**Before**

image

**After**

image


NickCraver:
Do you have some flexbox shenanigans happening in the page? I can't repro this, so trying to see what's happening before making changes.

xt0rted:
I'm using bootstrap 3 with the bootstrap 4 grid, so there very well might be. I'll take a better look into this later tonight.

NickCraver:
Alrighty I'll try some display: flex and try to break things here soon as time allows. On the project running: yeah, you'll need IIS setup there, see #184 for details. I'm hoping this gets fixed in https://github.com/dotnet/project-system/issues/2670 that I filed a while back. I wouldn't hold my breath there, though.

NickCraver:
@xt0rted can you try latest master? I went with a different fix that aligns with how the other stuff works (top level instead of per-element) in 7b1052da962f87cac9366a873b4bbe7b82055864. I broke/fixed it locally and I think this should resolve it.

Turns out we're seeing this on Stack Overflow as well...I just hadn't expanded any trivial gaps :) Thanks for the report and the help!

xt0rted:
@NickCraver just tried 4.0.115 off myget and both this and the double query count look to be fixed.

xt0rted:
I forgot to mention that when getting started with this project I needed to install the [Web Compiler](https://marketplace.visualstudio.com/items?itemName=MadsKristensen.WebCompiler) extension in order to get the less to compile. There wasn't mention of this in the readme so it might be helpful for future contributors to know that.

The recent versions on MyGet aren't marked as betas anymore. Wanted to say something in case they should be.

NickCraver:
Yep, intentional release number since they're stable, and good call on the README...added in 059b90e53f141b78e624e46b70f6db3d22c016fb.

I really appreciate all your help, makes me so happy for people to find this useful and then to go beyond that with help like this. <3

PR #289, opened 2018-06-07
Fixed broken generation of CSS classes on mp-result element

s08cz1:
I stumbled upon this problem after upgrading to latest version. CSS classes are being joined without any separators what brakes it a bit due to JS not being able to select right elements. I fixed it by adding default space prior to any new CSS class being added.

image


NickCraver:
Good catch, thanks!

PR #297, opened 2018-06-15
Map missing data types in SQL formatter

StuffOfInterest:
Add in translations for missing DbType enumeration values.


NickCraver:
Looking great, thanks!

PR #301, opened 2018-06-26
'profile' tag helper enhancement

StuffOfInterest:
Removes 'profile' tag from HTML page output and doesn't wrap code in 'using' statement when MiniProfiler is inactive to boost rendering performance.

Sorry for the five commits in the pull request.  Had issues resyncing my development clone.  I guess squashing the request will take care of the noise.

As I'm implementing MiniProfiler in a project at work I'm running into things like this that I'm trying to correct.  I've been using MiniProfiler for over seven years now, so glad to give a little bit back.

### DIFF
--- a/benchmarks/MiniProfiler.Benchmarks/Benchmarks/JSONLibBenchmarks.cs
+++ b/benchmarks/MiniProfiler.Benchmarks/Benchmarks/JSONLibBenchmarks.cs
@@ -28,29 +28,37 @@ namespace Benchmarks
 
         [Benchmark(Description = "Serialize: Simple MiniProfiler (Newtonsoft)")]
         public string SimpleSerializeNewtonsoft() => JsonConvert.SerializeObject(_simpleProfiler);
+
         [Benchmark(Description = "Serialize: Simple MiniProfiler (Jil)")]
         public string SimpleSerializeJil() => JSON.Serialize(_simpleProfiler, JilOptions);
+
         [Benchmark(Description = "Serialize: Simple MiniProfiler (ServiceStack)")]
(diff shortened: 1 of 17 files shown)


NickCraver:
I'm not sure I follow the changes here, can you help me understand? The only net change I see is the `using()` overall, but that's already a `using (null)` when MiniProfiler is off (this is a global optimization, and doesn't double-check `.Current` - that's why `.Step()` is an extension that handles nulls). That specific change isn't necessary...and should really be less optimal than current.

The `<profile>` tag in the page though: I don't see that in the commit. Is this an issue, is it being rendered in the page somewhere? If that's the case: yes, need to address that for sure.

StuffOfInterest:
Setting `output.TagName = null;` causes the `<profile>` tag to be removed from the output HTML.  If it wasn't for all the using() related code it would have been a one liner.

As for the using change, I knew that the result was going to be a `using(null)` when MiniProfiler is disabled, but I wasn't sure how that would compile below the surface.  Is the wrapping code for a using with no lock object as performant as doing the null check?  If so, then there is no need for that part of the change. I was just hoping to minimize the impact of the MiniProfiler code when it is disabled.

NickCraver:
Ahhh, missed the line! Indeed that change is necessary (thanks for the catch!)

The `using` changes aren't necessary though - can you please move `output.TagName = null;` to top of the method and remove the `using` `if`/`else` changes? Then happy to merge in :)

StuffOfInterest:
Done.

Not sure I'll ever be used to how git updates pull requests automatically with further checkins to a branch.

NickCraver:
Merged in, thanks for the contribution here! I completely missed that was happening :)

Issue #302, opened 2018-06-26
JavaScript timing in MiniProfiler results


StuffOfInterest:
I've seen scattered references but nothing concrete regarding capturing timing information in JavaScript code and having it displayed in the MiniProfiler results.  I'm dealing with issues where the DOM load is taking longer than the service side execute and render for pages.  The page uses several script libraries so it would be nice to capture timing information on script execution with something like:

```
window.MiniProfiler.StepIn("step name");
SomeLongRunningCode();
window.MiniProfiler.StepOut();
```

or maybe:

```
window.MiniProfiler.Step("step name", function() {
   SomeLongRunningCode();
});
```

Being that there are cases where loading the script needs to be timed, the former syntax may be more versatile.  That version would also be easier to suppress if MiniProfiler is not active (perhaps with a tag helper in cshtml).

NickCraver:
You can indeed do this today, you'll want to init the timing script up top, e.g.:
```cshtml
@this.InitClientTimings()
```
And your code would look like:
```js
if (window.mPt) mPt.start("step name");
SomeLongRunningCode();
if (window.mPt) mPt.end("step name");
```
Or, there are helpers, like this:
```cshtml
@this.TimeScript("step name", @<script>SomeLongRunningCode();</script>)
```

You can also search the sample projects for `ClientTimingHelper` to see some examples of it in play.

StuffOfInterest:
One small problem with this.  It lives in the MiniProfiler.Mvc5 project.  I'm working on an ASP.NET Core app.  That explains why I wasn't able to make the `InitClientTimings()` extension method work before.

Being that the core code for the helper is in a shared project (MiniProfiler.Shared) and the code for the extension methods is pretty straight forward, it shouldn't take too much to turn this into a couple of tag helpers for .NET Core.  I'll see what I can do there when I have some time available.  For now, I may just hack it into a project I'm working on to get a look at it.

StuffOfInterest:
Aside from me being selectively blind this early in the morning and not seeing the new entry right away, it works. 

First, I put ```@Html.Raw(StackExchange.Profiling.ClientTimingHelper.InitScript)``` just below the ```<mini-profiler />``` at the top of the page body.  Next, I wrapped the part of the page which loads a bunch of other scripts with ```<script>mPt.start('Layout script load')</script>``` and ```<script>mPt.end('Layout script load')</script>```.  

The page loads and when I bring up the MiniProfiler output I see this:

[...]

PR #304, opened 2018-06-29
Tag helpers for client timing scripts

StuffOfInterest:
Two tag helpers to initialize client timing scripts and to wrap a section of code for timing.  The initialization tag is optional as it will be pre-pended to the timing script if it hasn't already been loaded.  Also, the initialization script will only render once in case someone accidentally adds it twice.

Code below is how I was testing the behavior for the helpers.

```
<mini-profiler-timing />
<mini-profiler-timing />
<div></div>
<profile name="Render JS delay">
	<profile-script name="JS delay">
		<script language="javascript">
			alert("Wait for a second");
		</script>
	</profile-script>
</profile>
```

This partially addresses issues I mentioned in issue #302.


NickCraver:
Sorry I'm doing lots of physical building this weekend so away a bit, but wanted to update: I don't want to blindly add the init script tag helper here. Or at least, by itself. A few thoughts that came to mind:
- If we add the helper, I find `<mini-profiler-timing>` to be very confusing on the name front (and it auto-completed with `<mini-profiler>` which is bad) - but need to figure out naming with item 2:
- I wanted to see if the first invocation of `<profile-script>` (or new name if we need both and align them) can be what includes the script *if it hasn't been* already (in this page render).

I need to see what mechanisms exist for the 2nd of "has this been called yet in the page". If we can do that, people need not ever use the manual script init as they must in MVC5 today, we can automatically include the script at the first usage. For anyone manually profiling some JS without helpers and script tags (e.g. in a .js file even), so...yeah I guess we'll need that first helper.

I hope with some tweaks here we can do more than just tag-helper-ify MVC5, I think we can drastically improve the usability by removing the need to know about the implementation from the user - make sense? Thoughts?

StuffOfInterest:
Yea, I need to lay slate stones for a small patio in 95 degree weather this weekend so I'm not going to be doing much code either.

Naming if of course easily adjustable.  I picked the names to keep a close association with the existing tag helpers.  Perhaps `<mini-profiler-clienttiming>` would be more descriptive?  Or, maybe just `<profiler-client-timing>`?  Reversing the name `<profile-script>` to `<script-profile>` may also be worth considering as it would avoid confusion with the `<profile>` tag and it would fit closer to script blocks in the HTML. 

On your second bullet, that is exactly what is done.  I left off a couple of letters in my pull request description which were later corrected.  You must have gone off of the original e-mail notification.  The beginning of the `<profile-script>` helper has this block of code:

```
if (!ViewContext.ViewData.ContainsKey(MiniProfilerTimingTagHelper.ClientTimingKey))
{
    pre.Append(ClientTimingHelper.InitScript);
    ViewContext.ViewData[MiniProfilerTimingTagHelper.ClientTimingKey] = true;
}
```

The ViewContext is a convenient way for the helpers to talk to each other and coordinate.  With the key in there you don't have to use the initialization helper if all you are going to do is put in tag helper blocks.

There is a use case where the `<mini-profiler-timing/>` helper is still needed.  If all you are going to do is embed `if(window.mPt)...` blocks in a JS file or inside of `<script>` blocks then there is no profiling tag to initialize the timing.

NickCraver:
@StuffOfInterest I agree that *should* work, but in cases where you're rendering the view separately (e.g. razor is connected now), we'd double init (we have this case at SO) - but I guess that's rare enough it's okay. I was seeing if there's something *more* global to use here...but thinking it through, yeah I agree this is good enough for the 99.9% case and the other we can address with a boolean arg (disabling init) if we need to later.

Okay so let's think through the behavior here - I'm back to unsure we need 2 helpers. For example, if we do some re-org of the logic in `<profile-script>` (e.g. we need not `StringBuilder` if not first). Can we detect that the tag helper is used as self-closing (without relying on empty child content, which is valid)? If so, `<profile-script />` would cause the init we want, without a second helper. Thoughts?

StuffOfInterest:
A tweak to the initialization script itself could deal with the 1% cases where it ends up being loaded twice.  Just make sure that it checks that `window.mPt` does not exist before creating it.

As for eliminating the other helper completely, I can think of two ways to do it.

First is just as you say.  Second would be to put a parameter on the `<mini-profiler>` tag to include the timing script and tell users to put it at the beginning of the page.  This would cut down on the total number of helpers that have to be put on the page.

I prefer your option for two reasons.  First, it keeps the logic for client timing separate from profile display.  Second, the extra parameter is more cumbersome as it requires being more specific about where the `<mini-profiler>` tag goes on the page.

I can take a crack at the logic for detecting a self closing tag in the next day or two.  Any feeling on which name is better for the one tag that will remain?

StuffOfInterest:
Refactored.  Now if you include this:

```
<profile-script />
<profile-script />
<div></div>
<profile name="Render JS delay">
	<profile-script name="JS delay">
		<script>
			alert("Wait for a second");
		</script>
	</profile-script>
</profile>
```

You get this:

image

And if you do this:

```
<profile name="Render JS delay">
	<profile-script name="JS delay">
		<script>
			alert("Wait for a second");
		</script>
	</profile-script>
	<profile-script name="JS delay 2">
		<script>
			alert("Wait for another second");
		</script>
	</profile-script>
</profile>
```

You get this:

image

StuffOfInterest:
I've been using the current version of the script tag helper (with HttpContext for synchronization) for the last week.  The helper is communicating between the layout page and the view page without issue.  One change I would consider making is to switch the synchronization key from being a static string to using a GUID stored in a static field in the helper.  This would avoid having anything blatant in the HttpContext Items collection.

NickCraver:
@StuffOfInterest Sorry I've been busy elsewhere for a bit - can we please revert the GUID change and then I think we're good to merge. The keys in context (or elsewhere) are intentionally prefixed so you know what they are. If someone dumps their context and says "...what is this random GUID?" that's far worse :) We have a thing, it's there, it has a purpose...no need to be shy or lie about it to anyone. In fact, it can help debugging.

StuffOfInterest:
Done.  Guess I was trying to be too clever there.  I just have a general tendency to hate magic strings in code.

PR #305, opened 2018-06-29
Make client timing bar more visible in results

StuffOfInterest:
Changes the bar for client timing results from almost white (which disappears on an LCD monitor) to a pale yellow.  Addresses one of the items in issue #302.

Example of the colors:

image


NickCraver:
I ummmmmm - how to put it delicately, okay I definitely agree there's a problem but I'll take a crack at some "alternative" color options :D

StuffOfInterest:
Well, I did say over in issue #302 that "I have no sense of style".

Any color change will help.  I just didn't want to go with grey, which is what happened when I lowered the color from `F1F1F1` to `C1C1C1` which is why I switched to `C1C170`.

I changed the _Layout.cshtml on the work I'm going to include an override for the color so it looks better locally for me now.  Ended up putting in something like this (going from memory):

```
<style>
    .mp-result table.mp-client-timings .t-unknown div {
        background: #C1C170 !important;
    }
</style>
```

Speaking of which, it might make sense to apply an empty or undefined style class to these elements so that the override syntax would be simpler.  It would be much easier for someone to override `.mp-result-bar-clienttiming` than `.mp-result table.mp-client-timings .t-unknown div`.  This would allow having a documented list of classes to override for someone who wants to customize their colors.

StuffOfInterest:
Changed by Nick in a separate commit.

Issue #313, opened 2018-07-31
Factory provided DataAdapter.Fill throws InvalidOperationException


oising:
So, here's the pseudo-code (runtime is 4.7.2634.0):

```
var factory = DbProviderFactories.GetFactory("System.Data.SqlClient")
var profiledFactory = new ProfiledDbProviderFactory(factory);
var connection = profiledFactory.CreateConnection();
connection.ConnectionString = "..."

// ... other code here ... sqlCommand is a SqlCommand instance, with no connection info.
// it is create elsewhere - i.e. not created from profiledFactory 

var command = new ProfiledDbCommand(sqlCommand, connection, MiniProfiler.Current);
connection.Open()

var adapter = profiledFactory.CreateDataAdapter()
adapter.SelectCommand = command;
var data = new DataSet();

// this throws: IOE - ExecuteReader: Connection property has not been initialized.
adapter.Fill(data); 
```
When debugging, I can see that the connection is open. The ProfiledDbCommand's Connection property is using the ProfiledDbConnection instance. but the call stack blows up in SqlCommand.ValidateCommand(). This is because ProfileDbCommand does not assign the passed connection to the wrapped SqlCommand. 

This is a very old code base, hence the DataSet and weird patterns. And hence the profiling efforts :)

```
   at System.Data.SqlClient.SqlCommand.ValidateCommand(String method, Boolean async)
   at System.Data.SqlClient.SqlCommand.RunExecuteReader(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, String method, TaskCompletionSource`1 completion, Int32 timeout, Task& task, Boolean& usedCache, Boolean asyncWrite, Boolean inRetry)
   at System.Data.SqlClient.SqlCommand.RunExecuteReader(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, String method)
   at System.Data.SqlClient.SqlCommand.ExecuteReader(CommandBehavior behavior, String method)
   at System.Data.SqlClient.SqlCommand.ExecuteDbDataReader(CommandBehavior behavior)
   at System.Data.Common.DbCommand.ExecuteReader(CommandBehavior behavior)
   at StackExchange.Profiling.Data.ProfiledDbCommand.ExecuteDbDataReader(CommandBehavior behavior)
   at System.Data.Common.DbCommand.System.Data.IDbCommand.ExecuteReader(CommandBehavior behavior)
[...]

PR #314, opened 2018-07-31
ProfiledDbCommand does not set connection for wrapped command

oising:
See: https://github.com/MiniProfiler/dotnet/issues/313

### DIFF
--- a/src/MiniProfiler.Shared/Data/ProfiledDbCommand.cs
+++ b/src/MiniProfiler.Shared/Data/ProfiledDbCommand.cs
@@ -54,7 +54,11 @@ namespace StackExchange.Profiling.Data
         {
             _command = command ?? throw new ArgumentNullException(nameof(command));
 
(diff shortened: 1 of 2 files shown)


rhubley:
Wouldn't the fix be to use the setter?

        public ProfiledDbCommand(DbCommand command, DbConnection connection, IDbProfiler profiler)
        {
            _command = command ?? throw new ArgumentNullException(nameof(command));
            DbConnection = connection;

            if (profiler != null)
            {
                _profiler = profiler;
            }
        }

oising:
No. Look harder :)

oising:
@rhubley Sorry, I should not be so deliberately obtuse. You're right, it would fix it if I used the setter, but what if someone overrides that property in a derived class? I wouldn't be calling the code I thought I was anymore, and the bug may get reintroduced in a different manner. Generally speaking, calling virtual members in constructors is a bad idea, unless the class is sealed.

PR #325, opened 2018-09-13
Added an option for a custom ProfiledDbDataReader implementation

kekekeks:
Some ADO providers have extended API for database-specific functions, a notable example is Npgsql.

Some ORMs (e. g. linq2db) are aware of said functions and [try to use](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLDataProvider.cs#L208) them. Fortunately linq2db doesn't hardcode any type references and discovers needed APIs via reflection, it even allows to [configure the expected reader type](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLDataProvider.cs#L341) when needed.

This PR adds an option to substitute the standard ProfiledDbDataReader with a derived type that can reimplement said database-specific API


NickCraver:
I see the intent here, but I think we can improve the approach. The overall issue I see is this creates a lot of complicating to overload *just one* of the types involved (Connection, Command, Reader, Transaction, Adapter, ProviderFactory). To have this expand to be extensible for everything would be quite a lot of code (and more than a little messy honestly).

The intent of MiniProfiler is to allow extensibility here, but the approach intent is inheritance (as MiniProfiler itself does). For example in this case could be accomplished like this:
```c#
public class MyProfiledConnection : ProfiledDbConnection
{
    public MyProfiledConnection(DbConnection connection, IDbProfiler profiler) : base(connection, profiler) { }
    protected override DbCommand CreateDbCommand() => new MyProfiledCommand(WrappedConnection.CreateCommand(), this, Profiler);
}

public class MyProfiledCommand : ProfiledDbCommand
{
    public MyProfiledCommand(DbCommand command, DbConnection connection, IDbProfiler profiler) : base(command, connection, profiler) { }
    protected override DbDataReader ExecuteDbDataReader(CommandBehavior behavior) => ...
}
// and so on...
```

It may be more code for a single user to have, but it's the only realistic way to override the whole tree and have something maintainable for all. Thoughts?

kekekeks:
Switched to `protected virtual` extension points

src/MiniProfiler.Shared/Data/ProfiledDbCommand.cs
```diff
@@ -196,6 +196,12 @@ public override UpdateRowSource UpdatedRowSource
             set => _command.UpdatedRowSource = value;
         }
 
+        /// <summary>
+        /// Creates a wrapper data reader for <see cref="ExecuteDbDataReader"/> and <see cref="ExecuteDbDataReaderAsync"/> />
+        /// </summary>
+        protected virtual DbDataReader CreateDataReader(DbDataReader original, IDbProfiler profiler)
```
NickCraver: Let's align the naming here - we have `CreateDbCommand` with Db elsewhere and `CreateDataReader` here - can you please change this to `CreateDbDataReader` (more correct given the return type).

PR #334, opened 2018-10-30
Handle Exception in DataReader (Fire OnError event)

brightness007:
Some SQL command may failed in DataReader. For example, following SQL script:

```
INSERT INTO [dbo].[Demo] ([Name]) OUTPUT INSERTED.ID VALUES ('ABCD')
INSERT INTO [dbo].[Demo] ([Name]) OUTPUT INSERTED.ID VALUES ('ABCD')
```
The second one will failed to execute since NAME column has an unique index on it.

This fix will handle this error and call OnError function to capture it.

### DIFF
--- a/src/MiniProfiler.Shared/Data/ProfiledDbCommand.cs
+++ b/src/MiniProfiler.Shared/Data/ProfiledDbCommand.cs
@@ -199,8 +199,8 @@ namespace StackExchange.Profiling.Data
         /// <summary>
         /// Creates a wrapper data reader for <see cref="ExecuteDbDataReader"/> and <see cref="ExecuteDbDataReaderAsync"/> />
         /// </summary>
-        protected virtual DbDataReader CreateDbDataReader(DbDataReader original, IDbProfiler profiler)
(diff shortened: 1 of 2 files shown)


NickCraver:
I understand the intent here (and appreciate the change!), however, it is a breaking change that we couldn't make until a vNext major release. A few things break here:
- Constructors have changed (though that can be mitigated by having new overloads)...but we have to assume commands could be null.
- Async behavior changes - we're not creating our own async state machine and wait infrastructure before returning on long readers. This can be really bad for streaming scenarios. This happens even in the no-profiler-active scenario. Here, with the new a	`await`, we would add overhead even when not profiling.

I'm not sure we *should* capture here, since an error is going to go boom, and the timings on "it took this long...but went boom" isn't as helpful. I think maybe this should be considered in a much more global context of "should MiniProfiler capture all errors?"...which I'd argue isn't the goal and purpose of a profiler.

What are your thoughts?

NickCraver:
Given no activity - going to close out based on the above.

Issue #337, opened 2018-11-09
Ignore global ContractResolver in JsonConvert.DefaultSettings


mortenholmgaard:
When we have this `JsonConvert.DefaultSettings` it make mini profiler fail. Please make an explicit `ContractResolver `for mini profiler so that ignores global `ContractResolver `settings. We will properly have to skip using the mini profiler, just because our entire applications is build with this settings:
```
JsonConvert.DefaultSettings = () => new JsonSerializerSettings
{
	ContractResolver = new CamelCasePropertyNamesContractResolver()
};
```
This problem is also mentioned in this issue: https://github.com/MiniProfiler/dotnet/issues/200

NickCraver:
It took me a while to stare at this and realize what's happening. Somewhere internally, and I'm not sure if this is new, Newtonsoft is coalescing *individual properties* of the settings to the defaults. So unless we specify everything that matters on our overrides (e.g. "here's 1 thing that's different...and the other 60 defaults...and hope you don't add any!")...we get bitten.

I really, really dislike this rollup behavior. It's terrible for any consumer depending on stability. But in latest, the `ContractResolver` for MiniProfiler is explicitly set to `DefaultContractResolver` now. It'll be up on MyGet soon, and after a sweep of all issues I'll do a NuGet update with all fixes as well.

Thanks for the report!

mortenholmgaard:
Great - when will you expect the NuGet update to go out?

robbihun:
Looks like there may be a few more settings here - i agree with @NickCraver, this coalescing of the settings is annoying...

In our code we have the following:
```
            JsonConvert.DefaultSettings = () => new JsonSerializerSettings
            {
                Formatting = Formatting.Indented,
                TypeNameHandling = TypeNameHandling.Objects,
                ContractResolver = new CamelCasePropertyNamesContractResolver()
            };
```

It seems `TypeNameHandling.Objects` causes the following error:
![image](https://user-images.githubusercontent.com/138997/50112669-6ee36880-0205-11e9-9158-53bb94ad9db8.png)

I submitted #349 for this...

PR #349, opened 2018-12-17
Update ExtensionMethods.cs

robbihun:
Added explicit setting for TypeNameHandling in the JsonSerializerSettings to prevent overridden defaults from causing breaking changes to the MiniProfiler. Related to #337


NickCraver:
Well this rabbit hole sucks :) Thanks for fixing!

Issue #364, opened 2019-02-19
Flash of unstyled content


programcsharp:
I'm getting a flash of unstyled content, particularly on Firefox:

![image](https://user-images.githubusercontent.com/994270/53016288-822fd180-340a-11e9-80b2-682f009d0829.png)

Skimming the code, it looks like the CSS is deferred which means there's a moment when the popup is rendered on the page but doesn't have style: https://github.com/MiniProfiler/dotnet/blob/6b3a42a65fed15b8f8b35804668fbc2ea429402b/src/MiniProfiler.Shared/ui/lib/MiniProfiler.ts#L350

An idea for a fix would be to stick in some default style in doInit() to hide the wrapper initially (before the deferInit):

 $('head').append(`<style type="text/css">.mp-results { display: none }</style>`);

And then in the includes.less, show the wrapper (since the style will be in place so it's hidden correctly by the other styles):

.mp-results { display: block }

If that solve is sufficient, let me know and I'll submit a PR.

Another thought would be to insert the css as a string into the .js -- gzipped it'd only add ~3KB to an already 47KB file and saving a network request is probably worth it since it's cached anyway... oh wait, it isn't. Is there a reason ProcessRequest doesn't slap a far futures expiry on there? It's got the version number in the URL so it should be a prime candidate.

NickCraver:
> Is there a reason ProcessRequest doesn't slap a far futures expiry on there?

Nope...this used to be the case due to lack of proper cache breakers long ago - now we have per-version breakers that are fine to cache for say a month. Overall, it's a regression for sure. It should be cached which makes this a 1-time load thing and not worth complicating further. I'll poke this issue when a new build is up with some cache headers.

programcsharp:
Couldn't that still lead to a flash? Looks like it's doing a 100ms setInterval on deferInit, so you could have content in place before the CSS, even with cache. Doing the initial CSS hide + show on real CSS load would ensure that doesn't happen.

NickCraver:
Could it happen the very first time a user ever loaded MiniProfiler (usually via the corner item, which isn't shown until the JS loads later)? Yes. But in practice, you won't notice that...because it happens before the first corner block shows, and then it's not an issue for a month.

programcsharp:
Well, the way I read:

[...]

PR #367, opened 2019-03-09
Fix the flash

programcsharp:
#364


NickCraver:
Not sure if the workaround works here (noted in #364), so closing this out to cleanup. I'd love to hear if the simpler workaround for Firefox works!

PR #369, opened 2019-03-15
Added PostgreSql storage

Douglas-Cleghorn:
This adds PostgreSqlStorage. It's mostly just a copy of the SqlServerStorage class.


tests/MiniProfiler.Tests/MiniProfiler.Tests.csproj
```diff
@@ -17,7 +17,13 @@
   </ItemGroup>
   <ItemGroup Condition="'$(TargetFramework)' == 'net461'">
     <PackageReference Include="sqlite" Version="3.13.0" />
+    <PackageReference Include="Npgsql" Version="4.0.5" />
     <ProjectReference Include="..\..\src\MiniProfiler.Providers.SqlServerCe\MiniProfiler.Providers.SqlServerCe.csproj" />
+    <ProjectReference Include="..\..\src\MiniProfiler.Providers.PostgreSql\MiniProfiler.Providers.PostgreSql.csproj" />
+  </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp2.0'">
+    <PackageReference Include="Npgsql" Version="4.0.5" />
```
NickCraver: Since it's in both - we can just move this reference to the unconditional group above.

global.json
```diff
@@ -2,4 +2,4 @@
   "sdk": {
     "version": "2.1.300"
   }
-}
\ No newline at end of file
+}
```
NickCraver: Can we please undo the diff here, to eliminate it from git history? A quick way to do that is `git checkout master -- global.json` and committing it - then that commit will zero out in the squash of the PR later :)

src/MiniProfiler.Providers.PostgreSql/PostgreSqlStorage.cs
```diff
@@ -0,0 +1,489 @@
...
+    /// Understands how to store a <see cref="MiniProfiler"/> to a PostgreSQL Server database.
+    /// </summary>
+    public class PostgreSqlStorage : DatabaseStorageBase
+    {
+        /// <summary>
+        /// The schema that the tables are stored in. 
+        /// </summary>
+        public readonly string Schema = null;
+
+
```
NickCraver: nit: extra line return

src/MiniProfiler.Providers.PostgreSql/PostgreSqlStorage.cs
```diff
@@ -0,0 +1,489 @@
...
+{
+    /// <summary>
+    /// Understands how to store a <see cref="MiniProfiler"/> to a PostgreSQL Server database.
+    /// </summary>
+    public class PostgreSqlStorage : DatabaseStorageBase
+    {
+        /// <summary>
+        /// The schema that the tables are stored in. 
+        /// </summary>
+        public readonly string Schema = null;
```
NickCraver: The pattern here is to pull in the table names *with schema* (check out MySql and SqlServer) - this simplifies everything down to the same constructor as the other providers and makes maintenance easier and full control explicit. The tables don't even *have* to be in the same schema. Or any schema. It can be relative to the current user.

In short: we should remove this and offer the same constructor patterns as the existing providers here.

src/MiniProfiler.Providers.PostgreSql/PostgreSqlStorage.cs
```diff
@@ -0,0 +1,489 @@
...
+            {
+                MiniProfilersTableWithSchema = MiniProfilersTable;
+                MiniProfilerTimingsTableWithSchema = MiniProfilerTimingsTable;
+                MiniProfilerClientTimingsTableWithSchema = MiniProfilerClientTimingsTable;
+            }
+        }
+
+        private string _saveSql;
+        private string SaveSql => _saveSql ?? (_saveSql = $@"
+INSERT INTO {MiniProfilersTableWithSchema}
```
NickCraver: On these - we'll just need to use the base fields, e.g. `MiniProfilersTableWithSchema` => `MiniProfilersTable`.

src/MiniProfiler.Providers.PostgreSql/PostgreSqlStorage.cs
```diff
@@ -0,0 +1,489 @@
...
+                MiniProfilersTableWithSchema = $"{Schema}.{MiniProfilersTable}";
+                MiniProfilerTimingsTableWithSchema = $"{Schema}.{MiniProfilerTimingsTable}";
+                MiniProfilerClientTimingsTableWithSchema = $"{Schema}.{MiniProfilerClientTimingsTable}";
+            }
+            else
+            {
+                MiniProfilersTableWithSchema = MiniProfilersTable;
+                MiniProfilerTimingsTableWithSchema = MiniProfilerTimingsTable;
+                MiniProfilerClientTimingsTableWithSchema = MiniProfilerClientTimingsTable;
+            }
```
NickCraver: This will be a moot point in context of removing this constructor, but in the `schema == null` case, `*WithSchema` is a misnomer - it's still a table without schema, relative to the logged in SQL user, etc.

tests/MiniProfiler.Tests/MiniProfiler.Tests.csproj
Douglas-Cleghorn: netcoreapp1.1 wasn't supported by the current version of Npgsql. I can add in an old version of Npgsql if it needs to work with 1.1.

Douglas-Cleghorn:
It looks like the build failed on a network timeout. The schema stuff is gone now. I hadn't thought of passing the schema in with the table name. That makes a lot more sense. 

Let me know if you need any more changes to the Test project.  Otherwise, I think the build just needs to re-run.

Issue #370, opened 2019-03-16
Uncaught TypeError: i.Started.toUTCString is not a function


tomglover24:
I was looking to integrate MiniProfiler to an ASP.net Core 2.2 MVC app and have come across a rendering issue. This issue only affects rendering of the MiniProfiler widget and seems to be related to the format that Started is being provided in JSON back to the widget.

**Setup**
- .Net Core 2.2
- ASP.Net Core 2.2.0 MVC
- MiniProfiler.AspNetCore.Mvc 4.0.165
- MiniProfiler.Providers.MySql 4.0.165
- Default config from the getting started page, with the exception of:
```c#
options.Storage = new MySqlStorage(Configuration["ConnectionStrings:DefaultConnection"]);
```

**The Full Error**
```
Uncaught TypeError: i.Started.toUTCString is not a function
    at n.renderProfiler (includes.min.js?v=4.0.165+g04f8ac4653:formatted:4790)
    at n.buttonShow (includes.min.js?v=4.0.165+g04f8ac4653:formatted:4800)
    at Object.success (includes.min.js?v=4.0.165+g04f8ac4653:formatted:4552)
    at l (includes.min.js?v=4.0.165+g04f8ac4653:formatted:1825)
    at Object.fireWith [as resolveWith] (includes.min.js?v=4.0.165+g04f8ac4653:formatted:1876)
    at b (includes.min.js?v=4.0.165+g04f8ac4653:formatted:3539)
    at XMLHttpRequest.<anonymous> (includes.min.js?v=4.0.165+g04f8ac4653:formatted:3757)
```

**Example JS Response**
```JavaScript
[...]

PR #371, opened 2019-03-16
Updated isoDate regex to make micro seconds optional

tomglover24:
Updated isoDate regex to make micro settings optional, fixing an possible issue with the lack of precision returned by MySQL.

Full details can be found in this issue #370 

This fix has been tested in isolation with a number of iso dates without micro seconds, the full repo hasn't been built or tested unfortunately.


NickCraver:
Thanks for the fix! Verified locally and agree this looks great!

PR #379, opened 2019-04-24
Adding support to ElasticSearch Storage on MiniProfiler.Net.

reinaldocoelho:
Please, can you add an ElasticSearch Storage to MiniProfiler.Net?


NickCraver:
Hey there! Thanks for the pitch here. I've spent many hours looking at this (unfortunately with some frustration fighting NEST), and have concluded that:
- Yes, we can add Elasticsearch storage (though synchronous methods are problematic - that's the largest hurdle)
- *But*, it wouldn't look much like the current approach.

I've started a spike to get Elasticsearch in, but it's unlikely we can do so from this PR. I tried to adapt things and make it work, but the overall issues are:
- NEST is very complicated and expensive (we're honestly only doing very simple things: get/delete/update). A much simpler HTTP approach would work, and not have the dependency and breaking changes issues the Elastic clients have (this basically causes a rewrite).
- Other providers throw errors when they have issues so they can be addressed, this approach writes to `Console.WriteLine`. If that were correct to do, we still  can't assume there's a `Console` at all where MiniProfiler runs (also causes a rewrite - the NEST client has opinions here).
- The documentation is missing from all public methods (we can fix this though).
- The provider bakes in default connection endpoints (we can fix this).
- The tests don't skip when there's no instance or index (we can fix this).

I have a from-master branch that adds elastic with raw REST calls, but the synchronous bits I'll have to ponder on, as `HttpClient` doesn't support this (and honestly, rightfully so). It's unlikely I'll get back to this super soon but wanted to give an update on where everything is here.

millusdk:
Hi Miniprofiler maintainers,

I can see that there has been no updates to this pull request in a long time, and therefore I was wondering if support was still planned for Elasticsearch as a provider?

Best regards,
Emil

NickCraver:
@millusdk The problems above unfortunately haven't shifted any - it's unlikely this will land before a .NET 5 time-frame if we can make it work then. We likely want to get off the NEST dependency as an endpoint though and do something a bit more raw here to handle sync and async angles.

millusdk:
@NickCraver Thank you very much for the update. I can perfectly understand the reasoning with waiting until .Net 5 lands before looking more into this, and I also understand the wish to get off NEST dependency

Issue #342, opened 2018-11-16
Document URLs?


ChristianSauer:
Hi, I spend an hour yesterday how to find the urls of Miniprofiler:
"/results-index"
"/results-list"
"/results"
Maybe you could add a quick explanation in the docs?
Or maybe add the urls under the root url (e.g. return a json element which contains these three urls in HATEOAS-style?)

NickCraver:
These absolutely should be documented. I'll try and get to it soon or a PR would also be awesome if anyone gets to it before.

KrisJohnstone:
@NickCraver this can be closed. I didn't reference the issue correctly in the PR to have it auto close. :(

PR #386, opened 2019-05-13
Exposed End Points Added to Documentation

KrisJohnstone:
Addressing Issue #342
 
Looked at the ProcessRequest method in MiniProfilerHandler. I added it to the overview page as the routes don't changed based on framework and thus should be on a shared page. Further, creating a separate page seemed unwarranted for the amount of information.

Any other content that should be covered or any routes that I missed?


KrisJohnstone:
Closes: #342

PR #400, opened 2019-07-06
Adding ASP.NET Core 3.0 Support

NickCraver:
This set of changes drops view profiling for the moment in ASP.NET Core 3.x since the types we were wrapping as "pubternal" are no longer exposed in 3.0. On the framework side, https://github.com/aspnet/AspNetCore/pull/11730 is in progress so we can use the diagnostics API to hopefully re-enable this for the 3.0 release.

Samples for ASP.NET Core 2.2 and 3.0 are broken into separate samples for clarity though they are mostly the same.

Overall changes:
- (For Linux): Fix Azure DevOps builds (note they're not successfully running all tests yet - some providers are skipped, but they're working!)
- (For Linux): Change SQL Formatters to always use `\n`
- (For Linux): Add `WindowsOnly` to `[Fact]` and `[Theory]` for skipping inapplicable tests on Linux
- (For Linux): Add `xunit.runner.json` to make things work on Linux/Mono (only copied to output and in effect there)
- (For Linux): Null ref fix for SQL CE errors (happens on Linux)
- (For .NET Core 3.0): Add a `netcoreapp3.0` build for `MiniProfiler.AspNet*` libs
- (For .NET Core 3.0): Exclude view profiling for now (this is what was breaking, ultimately)
- (For .NET Core 3.0): Split ASP.NET Core v2 and v3 samples into different projects, since some fundamentals have changed
- (For all): Bump version to `4.1.0-preview.*` to reflect the preview dependencies for `netcoreapp3.0` builds
- (For all): Dropped `netcoreapp1.1` from testing (note: `netstandard1.5` is still built, it's just not tested)


sebastienros:
Any way we could get these changes on nuget, like before Monday EOD ? We are trying to ship Orchard Core right after dotnet 3.0, and right now we are using it on myget but we'd have to drop it as a default module if it's not on nuget. We just found out it was broken for us o. 3.0.

If there is something we can do, like testing, just let us know. And it's not critical for us either, just nice to have it oob if possible.

Thanks.

NickCraver:
@sebastienros if you can poke me as soon as 3.0 packages are available to depend on, I'll get a change and build up ASAP. We're running these preview builds already with no issues. Does the preview being on NuGet help for 3.0, ahead of stable? (if so I can do that shortly)

sebastienros:
The preview is fine for us, it's just about fixing the feature and it being available on nuget that matters. Thanks a lot.

NickCraver:
@sebastienros awesome - I just made final tweaks and tests on cleanup - `4.1.0-preview.32` is validating on NuGet now and should be available soon. Please shout if you have any issues!

PR #416, opened 2019-09-01
GitHub Actions Build

NickCraver:
Trying to get the build setup in actions here


.github/workflows/main.yml
```diff
@@ -0,0 +1,64 @@
+name: Main Build
+
+on:
+  pull_request:
+  push:
+    branches:
+    - master
+    paths:
+    - '*'
+    - '!/docs/*' # Don't run workflow when files are only in the /docs directory
```
NickCraver:
```suggestion
    branches2:
    - master2
    paths:
    - '*'
    - '!/docs/*' # Don't run workflow when files are only in the /docs directory
```

PR #426, opened 2019-09-23
Upgrade to 3.0 RTM packages and SDKs

NickCraver:
This bumps versioning off pre-release packages for the .NET Core 3.0 release today.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
     "version": "3.0.100",
-    "rollForward": "latestFeature"
+    "rollForward": "latestMajor"
   }
 }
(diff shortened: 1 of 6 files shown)


global.json
```diff
@@ -1,5 +1,6 @@
 {
   "sdk": {
-    "version": "3.0.100-preview9-014004"
+    "version": "3.0.100",
+    "rollForward": "latestFeature"
```
ctyar: Does the `latestFeature` is a valid value here?
The docs only have these values:
`Disable`, `LatestPatch`, `Minor`, `LatestMinor`, `Major`, `LatestMajor`

NickCraver: Good question! I'm honestly not sure - see https://github.com/dotnet/designs/pull/71/files#examples-of-usage which is still unclear. I'm okay with `latestMajor` anyway (and we can tweak this any time) - so let's go with that :)

Note the docs for CLI are not the same as `global.json` :)

PR #431, opened 2019-10-21
Open the connection once

bgrainger:
If the connection isn't already open, each call to `Execute(Async)` will open and close it, which is less efficient than opening it once and reusing the open connection.


NickCraver:
Looking good, thanks!

PR #432, opened 2019-10-21
Update MySqlConnector to 0.60.1

bgrainger:
MySqlConnector 0.30.0 is quite old; it seems like it would be best to use the latest version, which has the latest bug fixes and performance enhancements.

(MySqlConnector 0.30.0 is the oldest strong-named version, so if you want to rely on the oldest possible version that is compatible with anything your users want to upgrade to, then I understand if you want to close this PR. I just think it would be best to install the latest good version by default.)

### DIFF
--- a/src/MiniProfiler.Providers.MySql/MiniProfiler.Providers.MySql.csproj
+++ b/src/MiniProfiler.Providers.MySql/MiniProfiler.Providers.MySql.csproj
@@ -10,6 +10,6 @@
   <ItemGroup>
     <ProjectReference Include="..\MiniProfiler.Shared\MiniProfiler.Shared.csproj" />
     <PackageReference Include="Dapper.StrongName" Version="1.50.2" />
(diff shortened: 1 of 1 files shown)


NickCraver:
Change looks good but let me visit this tonight and see about the strong naming and where things are. Are the plans to strong name future versions?

bgrainger:
Sorry for being unclear.

The first version of MySqlConnector that was strong-named was 0.30.0. The latest (0.60.2) continues to be strong-named. I was trying to say that if you wanted to stick with the lowest-possible version number (for broadest compatibility...?), then 0.30.0 is good. But 0.60.2 hasn't dropped any TFMs, so I don't think there's really any reason not to upgrade.

Also, I just shipped a bugfix so I bumped this branch to 0.60.2.

NickCraver:
Pulling in - thanks again! :) You're a saint for maintaining so much around this driver <3

PR #442, opened 2019-11-30
Library Changes: Removing jQuery

NickCraver:
This removes the jQuery bundling and dependency, taking the raw size from 133KB down to 48.3KB.

Upsides:
- Smaller payload
- Fewer dependencies

Downsides:
- More maintenance/direct code in vailla JS/TS
- No IE <= 11 Support (I'll try this next without fetch and see what the diff is)

This also upgrades the TypeScript compiler and minifier which wasn't handling ES6 well. But this is a work-in-progress. Committing here because this is a stable and working ES6 version.


src/MiniProfiler.Shared/ui/lib/MiniProfiler.ts
```diff
@@ -422,27 +433,29 @@ namespace StackExchange.Profiling {
...
+
+                fetch(this.options.path + 'results', {
+                    method: 'POST',
+                    body: JSON.stringify(request),
+                    headers: {
+                        'Accept': 'application/json',
+                        'Content-Type': 'application/json'
+                    }
+                })
+                    .then(data => data.text())
```
xPaw: You can use `data.json()` 

https://developer.mozilla.org/en-US/docs/Web/API/Body/json

NickCraver: If you look at the next line, I'm not doing _exactly_ a `.json()` (as it's interpreting dates) like on line 390. Is there an overload that takes the custom arg? I didn't find one looking but awesome if there is!

xPaw: Oh I didn't notice that. I would just parse the date after json is parsed. Any reason you need to use the second arg of JSON.parse?

`json.Started = new Date(json.Started);`

NickCraver: Yeah, to parse the dates :) No need to allocate twice here (and change types). Note: timings are a tree - traversing that tree to parse all dates would be non-trivial in code and cost...best to do it while deserializing.

xPaw: Makes sense then, don't mind me.

src/MiniProfiler.Shared/ui/lib/MiniProfiler.ts
```diff
@@ -886,125 +899,302 @@ namespace StackExchange.Profiling {
...
-            // ensure they're in view
-            whatToScroll.scrollTop(whatToScroll.scrollTop() + rows.position().top - 100);
+        private scrollToQuery = (link: HTMLElement, queries: HTMLElement, whatToScroll: HTMLElement) => {
+            const id = link.closest('tr').dataset['timingId'];
+            const rows = queries.querySelectorAll('tr[data-timing-id="' + id + '"]');
+            rows.forEach(n => n.classList.add('highlight'));
+            const first = rows && rows[0];
+            if (first) {
+                // ensure they're in view
+                whatToScroll.scrollTop = whatToScroll.scrollTop + first.getBoundingClientRect().top - 100;
```
xPaw: You can look into https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

NickCraver: Oh awesome, TIL - thanks! I'll test this out tonight.

src/MiniProfiler.Shared/ui/lib/MiniProfiler.ts
```diff
@@ -886,125 +899,302 @@ namespace StackExchange.Profiling {
...
-                                if (bg.is(':visible')) {
-                                    bg.remove();
-                                } else {
-                                    active.removeClass('active');
+                        queries.querySelectorAll('pre code').forEach(block => mp.highlight(block as HTMLElement));
+                        return;
+                    }
+                });
+                // Background and esc binding to close popups
+                const tryCloseActive = (event: MouseEvent | KeyboardEvent) => {
```
ArtOfCode-: Other code (including new code in this PR) uses `function`; this uses `() => {}`. Probably best to standardise on one.

Does this get transpiled to ES5 before it gets to browsers? If not, ES6 shorthand syntax loses you IE support.

You may know all of this already, of course :)

src/MiniProfiler.Shared/ui/lib/MiniProfiler.ts
NickCraver: Done! https://github.com/MiniProfiler/dotnet/pull/442/commits/49aa25b5193a417d5048e1fef6d22fba246e3503

src/MiniProfiler.Shared/ui/lib/MiniProfiler.ts
NickCraver: A fair point! On this one, it should be `function tryCloseActive()` for consistency.

On why I'm not using a specific one: arrow functions don't bind `this` in the same way and we want to alter the element in the handler, e.g. `this.classList` - so I standardize on `function X()` for event handlers, and arrow functions everywhere else. Making that adjustment here :)

NickCraver: Took a look - normalized the others but we're in a block here and we can't declare it function style (and don't need `this`), so leaving as is for this case

src/MiniProfiler.Shared/ui/lib/MiniProfiler.ts
```diff
@@ -1,5 +1,4 @@
-﻿/// <reference path="./node_modules/@types/jquery/index.d.ts">
-/// <reference path="./node_modules/@types/extjs/index.d.ts">
+﻿/// <reference path="./node_modules/@types/extjs/index.d.ts">
```
NickCraver: Test Comment for a thing

NickCraver: Test Comment for debug

NickCraver: Comment 3

Issue #366, opened 2019-03-05
MiniProfiler.Providers.Redis target StackExchange.Redis


pergardebrink:
I see that the MiniProfiler.Providers.Redis is having a dependency on StackExchange.Redis.StrongName >=1.2.6. 

Is there a plan to change the dependency to StackExchange.Redis, now that the 2.0 version is strongly signed? Or will that not happen until MiniProfiler 5.x (maybe considered a breaking change)?

NickCraver:
Talked to @mgravell about this - my plan is to wait until we're doing Dapper v2 (since this is a breaking change) and eat the `StackExchange.Redis.StrongName` v1.x => `StackExchange.Redis` v2.x and `Dapper.StrongName` v1.x => `Dapper` v2.x at the same time. In both cases, we're dropping the .StrongName assembly and strong naming the main library to simplify things for everyone and eliminate a lot of pain and questions around the whole deal. After that, yeah it'll be a 5.x release, and we can rack up any other breaking changes plan in the mean time.

prat0088:
Any idea when this will happen?

Would aliasing StackExchange Redis 1.2.6 solve the problem?  That's what I'm doing now with my own project that is using 2.x along side MiniProfiler's redis storage.  It's annoying because Visual Studio isn't smart enough to figure out the alias so code completion breaks for StackExchange.Redis.  If we aliased StackExchange.Redis inside MiniProfiler.Providers.Redis then I wonder if it would resolve the issue.  Just thinking aloud since I don't have time to try it out this week.

NickCraver:
We'll need to do this in MiniProfiler v5 for safety's sake - at which point we'll drop < netstandard2.0 or 2.1 support overall. I expect this to happen in the next few months.

NickCraver:
Decided to put this in the 4.2.x release, see #473 for details!

Caspeco-MarcusOtterstrom:
Forgive me if this is the wrong place to ask, but is there a planned release date for 4.2.x? And will it target .NET Standard 2.0? We're looking forward to this dependency change. Right now our solution is to copy the files for `MiniProfiler.Providers.Redis` manually, since we can't install both `StackExchange.Redis.StrongName` and `StackExchange.Redis`.

Thank you! :)

NickCraver:
@Caspeco-MarcusOtterstrom it went up to NuGet this morning, was just finding the time after Stack Overflow dogfooding - 4.2.1 is now on NuGet :)

PR #453, opened 2020-01-28
Update StackExchange.Redis to latest

allyn-psyonix:
StackExchange.Redis no longer supports the old "Strongname" library that the dotnet MiniProfiler uses. The latest StackExchange.Redis is now the strongly named version by default. All updates since v2 are now only being applied to the base lib, StackExchange.Redis.

You may have reasons for not upgrading the Redis lib, but I wanted to propose an update. In particular, StackExchange.Redis.StrongName has been deprecated and is no longer receiving updates. The latest StackExchange.Redis is now strongly named by default. There are other changes that come along with the v2.x update that may affect your decision as well. Release notes are linked. v2.0.495 forward.

In particular, the v2 update includes this warning (emphasis mine):
`HARD BREAK: the package identity has changed; instead of StackExchange.Redis (not strong-named) and StackExchange.Redis.StrongName (strong-named), we are now only releasing StackExchange.Redis (strong-named). **This is a binary breaking change that requires consumers to be re-compiled**; it cannot be applied via binding-redirects`

https://stackexchange.github.io/StackExchange.Redis/ReleaseNotes.html

### DIFF
--- a/src/MiniProfiler.Providers.Redis/MiniProfiler.Providers.Redis.csproj
+++ b/src/MiniProfiler.Providers.Redis/MiniProfiler.Providers.Redis.csproj
@@ -8,8 +8,8 @@
     <TargetFrameworks>net461;netstandard2.0</TargetFrameworks>
   </PropertyGroup>
   <ItemGroup>
+    <PackageReference Include="StackExchange.Redis" Version="2.0.601" />
     <ProjectReference Include="..\MiniProfiler.Shared\MiniProfiler.Shared.csproj" />
-    <PackageReference Include="StackExchange.Redis.StrongName" Version="1.2.6" />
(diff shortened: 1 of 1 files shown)


NickCraver:
Indeed there are reasons :) See https://github.com/MiniProfiler/dotnet/issues/366 for details, the short version is: we need to do this on a major version (v5) release because the way transitive dependencies work we'd be forcing a Major redis upgrade on a minor or patch MiniProfiler release. It's just not a friendly thing to do :)

In case it helps context: I'm also one of the 2 maintainers of the StackExchange.Redis lib ;)

allyn-psyonix:
No worries, I figured it was something like that. Thanks for responding.

Issue #393, opened 2019-06-12
Ability to pass JavaScript nonce via RenderIncludes()


i4rilu:
Please implement an ability to pass **_nonce_** (random string - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script) via MiniProfiler.Current.RenderIncludes() method.

This would allow to apply Content-Security-Policy rules on the website and white-list MiniProfiler _includes.min.js_ JavaScript as a safe one.

NickCraver:
Yeah we could add this - just so I'm clear - is this asking for a `MiniProfiler.Current?.RenderIncludes(Nonce: "mynonce")` setup? If not, could you clarify with an example.

I'm also what @vcsjones thinks on adoption here :)

i4rilu:
Yes, that's correct.
Additional parameter for MiniProfilerWebExtensions.RenderIncludes() extension method.

vcsjones:
That seems useful. Would need to make sure that there is no caching of the script tags with the nonce. Perhaps another option to think about would be a factory somewhere, maybe on `MiniProfilerBaseOptions` that looks something like this:

```csharp
public Func<HttpContext, string> CspNonceFactory { get; set; }
```

So that you only need to configure it once. It takes an `HttpContext` since there needs to be some coordination of the nonce between the script tag and whatever renders the CSP header. Just thinking out-loud here though.

>I'm also what @vcsjones thinks on adoption here :)

Nonces are mostly well supported by everything except IE. Edge (the old Edge, not Edgium) does not support nonces on scripts that use `src`, only on inline scripts, but I think that's OK here.

NickCraver:
I'm taking a peek at this now - it'd have to be on the per-provider options which makes things a bit more complicated. That's because `HttpContext` differs between ASP.NET and ASP.NET Core and `Render` lives in the shared lib. But, adding a parameter is also a breaking change...so we'll need some overloads here to handle it. I'm debating if we're getting too many parameters in there, but allocating an object every time isn't great either. Going to ponder options here for a hot second before adding more overloads (but that's likely most practical at the moment).

NickCraver:
Hey all, sorry this lingered - how to version the `RenderIncludes` API was the vexing piece and I needed to consider a few more additions to it and what things collectively looked like.

[...]

PR #465, opened 2020-04-04
Add a nonce option and Render testing

NickCraver:
Fix for #393, allows passing a nonce through the new `RenderOptions` API added in #451.

Also a minor optimization for async...we don't need that attribute value for any browser that'll support us today.

### DIFF
--- a/samples/Samples.AspNetCore3/Views/Shared/_Layout.cshtml
+++ b/samples/Samples.AspNetCore3/Views/Shared/_Layout.cshtml
@@ -46,7 +46,7 @@
     @RenderSection("scripts", required: false)
 
     @* Simple options are exposed...or make a full options class for customizing. *@
-    <mini-profiler position="@RenderPosition.Right" max-traces="5" color-scheme="ColorScheme.Auto" />
+    <mini-profiler position="@RenderPosition.Right" max-traces="5" color-scheme="ColorScheme.Auto" nonce="45" />
     @*<mini-profiler options="new RenderOptions { Position = RenderPosition.Right, MaxTracesToShow = 5, ColorScheme = ColorScheme.Auto }" />*@
 </body>
 </html>
--- a/src/MiniProfiler.Shared/Internal/Render.cs
+++ b/src/MiniProfiler.Shared/Internal/Render.cs
@@ -84,7 +84,7 @@ namespace StackExchange.Profiling.Internal
             }
             if (renderOptions?.Nonce.HasValue() ?? false)
             {
-                sb.Append(" nonce=\"").Append(renderOptions.Nonce).Append("\"");
+                sb.Append(" nonce=\"").Append(System.Web.HttpUtility.HtmlAttributeEncode(renderOptions.Nonce)).Append("\"");
             }
 
             sb.Append(" data-max-traces=\"");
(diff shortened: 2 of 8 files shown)


samples/Samples.AspNetCore3/Views/Shared/_Layout.cshtml
```diff
@@ -46,7 +46,7 @@
     @RenderSection("scripts", required: false)
 
     @* Simple options are exposed...or make a full options class for customizing. *@
-    <mini-profiler position="@RenderPosition.Right" max-traces="5" color-scheme="ColorScheme.Auto" />
+    <mini-profiler position="@RenderPosition.Right" max-traces="5" color-scheme="ColorScheme.Auto" nonce="45" />
```
vcsjones, line 49: Given that nonce is supposed to be random on every page load, I'm not sure it makes sense to have it as an attribute on `mini-profiler`? Would the intention instead be to use something like `@NonceSource.GetNonce()`? 

As far as documentation / samples go, what is the story for "I want to inform miniprofiler about a nonce that comes from some piece of middleware?" The middleware shoves it in the context somewhere and it gets pulled in here?

src/MiniProfiler.Shared/Internal/Render.cs
```diff
@@ -82,6 +82,10 @@ public static class Render
             {
                 sb.Append(" data-start-hidden=\"true\"");
             }
+            if (renderOptions?.Nonce.HasValue() ?? false)
+            {
+                sb.Append(" nonce=\"").Append(renderOptions.Nonce).Append("\"");
```
vcsjones, line 87: Since I see that `sb` contains other markup elements, (line 114) it is presumed that it's contents are `HtmlSafe`? Is there any concern that `Nonce` is an XSS point, or at least requires some escaping?

Alternatively, what about simply rendering the incorrect thing? If a nonce is supposed to be random and someone's nonce generator uses the printable ascii range (0x20-0x7F), `&`, `"`, etc. could feasibly end up breaking rendering if this isn't escaping like I am concerned about.

It is assumed that the caller will do the escaping for miniprofiler

samples/Samples.AspNetCore3/Views/Shared/_Layout.cshtml
NickCraver: Correct - you'd pass in a unique value. Even putting say a `<Func<HttpContext, string>` is complicated due to the different primitives with .NET vs. .NET Core...maybe when we can drop .NET Full Framework support one day that gets easier since there's no common ancestor need.

src/MiniProfiler.Shared/Internal/Render.cs
NickCraver: Doh, absolutely - this was the first text thing a user can pass and definitely needs encoding, total goof. Updated (and added a few tests for these cases).

samples/Samples.AspNetCore3/Views/Shared/_Layout.cshtml
NickCraver: Oh sorry misread the second part - there's no built-in middleware story I can think of that makes sense here. The problem is generally you want MiniProfiler to profile the things you're doing and we have an order inversion problems with dependencies about the middleware doing it. Either direction we choose would be wrong for some subset of users. At least in v1, I'd want to leave it up to the users to handle their nonce generation.

NickCraver:
@vcsjones if time allows, mind taking one more peek please?

NickCraver:
@vcsjones Thanks!

Issue #469, opened 2020-04-16
Middleware Pipeline Branching can cause MiniProfiler to register diagnostic listeners multiple times.


haacked:
Suppose you want to use asp.net core [middleware pipeline branching](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/middleware/?view=aspnetcore-3.1#branch-the-middleware-pipeline).

Here's a contrived example.

I've registered MiniProfiler like so...

```csharp
services.AddMiniProfiler()
               .AddEntityFramework();
```

And I want to use it in two pipeline branches like so...

```csharp
app.UseWhen(httpContext => httpContext.Request.Query.ContainsKey("profile"),
    builder => 
        builder.UseMiniProfiler());

app.UseWhen(httpContext => httpContext.Request.Query.ContainsKey("observe"),
    builder => 
        builder.UseMiniProfiler());
```

In setting up this pipeline, `UseMiniProfiler` is called twice.

The `UseMiniProfiler` method [does this](https://github.com/MiniProfiler/dotnet/blob/6b3a42a65fed15b8f8b35804668fbc2ea429402b/src/MiniProfiler.AspNetCore/MiniProfilerBuilderExtensions.cs#L23-L25)...

```csharp
var listeners = builder.ApplicationServices.GetServices<IMiniProfilerDiagnosticListener>();
var initializer = new DiagnosticInitializer(listeners);
initializer.Start();
```

Which ends up calling this static twice...

```csharp
 DiagnosticListener.AllListeners.Subscribe(this)
```

So when building the middleware pipeline, we end up subscribing two instances of the EF listener to diagnostics. Thus every query gets reported twice by MiniProfiler. If I set up another branch, I get an additional query report for each query.

My workaround was to make sure I only call `UseMiniProfiler` in one branch, and called `UseMiddleware<MiniProfilerMiddleware>` in the others. It's a Haack, but it works. I figured I'd log this in case you had a better long term solution.

NickCraver:
It'll go through a Stack Overflow for dogfooding before NuGet (as always), but if you want a package with the fix, `4.2.0-preview.5` packages are up on MyGet now! Link: https://www.myget.org/feed/miniprofiler/package/nuget/MiniProfiler.AspNetCore.Mvc

PR #470, opened 2020-04-16
Shuffle DiagnosticListener to prevent double init

NickCraver:
Proposed fix for #469, this follows the intended pattern a bit closer and also makes `DiagnosticListener` accessible.

cc @haacked

### DIFF
--- a/src/MiniProfiler.Shared/DiagnosticInitializer.cs
+++ b/src/MiniProfiler.Shared/DiagnosticInitializer.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Threading;
 
 namespace StackExchange.Profiling.Internal
 {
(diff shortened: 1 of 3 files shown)


src/MiniProfiler.Shared/DiagnosticInitializer.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Threading;
```
haacked, line 4: Why this using statement?

NickCraver: Because I almost made it an `Interlocked` `int` swap :) (removing!)

haacked: Ha! I figured. Since this should be called in startup, the fact that it's not thread should be fine. You and I both know there will be someone out there who will try to parallelize app startup and they'll be wrong, but not sure it makes sense to worry about that. 👍

Issue #322, opened 2018-09-04
WrappedView breaks access to RazorView in ViewLocationExpander


PTwr:
In custom `IViewLocationExpander` I need to access `RazorView` from `(ViewLocationExpander.ActionContext as ViewContext).View as RazorView`. 

For now I "fixed" it with Reflection, `.GetType()?.GetField("_wrapped", BindingFlags.Instance | BindingFlags.NonPublic)`, but some civilized way would be nice.

`WrappedView` is currently `Internal` and wrapped `IView` is private. So I tried to provide my own implementation. However it must be provided via my own `ProfilingViewEngine` (which is public while sitting in `Internal` namespace), which in turn is provided in `Configure` hidden in internal classes underneath innocent looking `AddMiniProfiler` so replacing viewEngine requires building configuration from scratch. Which made reflection look like reasonable choice.

Is this blocked by design or for no apparent reason?

NickCraver:
You can see #162 for some details here. It's not intentionally blocked and I'd rather we not have this layer at all, but the current state of DiagnosticSource in ASP.NET MVC (no concrete or exposed types) is a nightmare to consume. I've asked (and gotten) the events fixed in https://github.com/aspnet/Mvc/issues/6222, but [the types are still anonymous and not exposed](https://github.com/aspnet/Mvc/blob/2.2.0-preview2/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Internal/MvcViewFeaturesDiagnosticSourceExtensions.cs#L23-L28). In short, they're terrible for consumption or profiling.

For context, I've discussed this quite a bit with the team that owns the diagnostic bits (though no one "owns" it overall...and that's the problem, IMO). Lots of context in the StackExchange.Redis discussion here: https://github.com/StackExchange/StackExchange.Redis/issues/833

[...]

PR #475, opened 2020-04-19
ASP.NET Core 3.x: use Diagnostic bits

### LINKED ISSUE #162: The types registered for MvcOptions and MvcViewOptions conflicts with other libraries.
AddMiniProfiler registers a type for MvcOptions and MvcViewOptions that can override or be overidden by another library.

Depending on what library is called first, the last registered type is used by MVC.

For example, FluentValidation does the same and conflicts with MiniProfiler, preventing ProfilingViewEngine from being used.
https://github.com/JeremySkinner/FluentValidation/blob/237ccfd581dcecf72209acdaabd4c41122fad98d/src/FluentValidation.AspNetCore/FluentValidationMvcExtensions.cs#L63

A workaround could be to register the ViewEngine using IServiceCollection.Configure()

```c#
services.Configure<MvcViewOptions>(options =>
{
    for (var i = 0; i < options.ViewEngines.Count; i++)
    {
        options.ViewEngines[i] = new ProfilingViewEngine(options.ViewEngines[i]);
    }
});
```

The same thing applies to MvcOptions which adds ProfilerActionFilter to the Filters collection. Here is my current workaround.

```c#
services.AddMvc(cfg =>
{
    cfg.Filters.Add(new ProfilingActionFilter());
})
```

NickCraver:
This adds timings based on the `DiagnosticListener` bits overhauled in https://github.com/dotnet/aspnetcore/pull/11730 with some further cleanup in https://github.com/dotnet/aspnetcore/commit/b23ea5b6683b08e8d168ccf49e0fc8515077ee2e

This would also resolve #162 and #322.

Entirely possible there are more efficient ways to do the correlation tracking (I hope!) - putting this up for more eyes.

Example profiling:
image

cc @benaadams @rynowak @vcsjones


src/MiniProfiler.AspNetCore.Mvc/MvcDiagnosticListener.cs
```diff
@@ -0,0 +1,166 @@
...
+        /// <param name="error">An object that provides additional information about the error.</param>
+        public void OnError(Exception error) => Trace.WriteLine(error);
+
+        private static string GetName(string label, ActionDescriptor descriptor)
+        {
+            var controller = descriptor.RouteValues.TryGetValue("controller", out var c) ? c : "UnknownController";
+            var action = descriptor.RouteValues.TryGetValue("action", out var a) ? a : "UnknownAction";
+
+            // TODO: Don't allocate this string more than once
+            return label + ": " + controller + "/" + action;
```
rynowak, line 48: Is there a reason to avoid using `ActionDescriptor.DisplayName`? This approach assumes everything is a controller, and possibly assumes that everything has a unique action/controller name (definitely not true).

ActionDescriptors all have a unique ID (not stable across launches) also

src/MiniProfiler.AspNetCore.Mvc/MvcDiagnosticListener.cs
```diff
@@ -0,0 +1,166 @@
...
+        }
+
+        private static string GetName(IActionResult result) => result switch
+        {
+            ViewResult vr => vr.ViewName.HasValue() ? "View: " + vr.ViewName : "ViewResult",
+            ContentResult cr => cr.ContentType.HasValue() ? "Content: " + cr.ContentType : "ContentResult",
+            // TODO: Other main ones?
```
rynowak, line 55: The main ones would be ObjectResult, StatusCodeResult, and JsonResult (legacy). 90% of the other action result types inherit from one of the first two.

src/MiniProfiler.AspNetCore.Mvc/MvcDiagnosticListener.cs
```diff
@@ -0,0 +1,166 @@
...
+            // PageFilterOnPageHandlerExecuted
+            BeforePageFilterOnPageHandlerExecutedEventData data => Start(data.Filter, "Filter (Execed): " + GetName(data.Filter)),
+            AfterPageFilterOnPageHandlerExecutedEventData data => Complete(data.Filter),
+
+            // PageFilterOnPageHandlerSelection
+            BeforePageFilterOnPageHandlerSelectionEventData data => Start(data.Filter, "Filter (Selection): " + GetName(data.Filter)),
+            AfterPageFilterOnPageHandlerSelectionEventData data => Complete(data.Filter),
+            // PageFilterOnPageHandlerSelected
+            BeforePageFilterOnPageHandlerSelectedEventData data => Start(data.Filter, "Filter (Selected): " + GetName(data.Filter)),
+            AfterPageFilterOnPageHandlerSelectedEventData data => Complete(data.Filter),
```
rynowak, line 161: How badly will things get messed up if the *end* event never fires due to an exception?

I'm not sure that we provide the guarantee that both events fire when an exception happens.

src/MiniProfiler.AspNetCore.Mvc/MvcDiagnosticListener.cs
NickCraver: Mainly just that it's so long, they'd all appear the same way in the UI (e.g. seeing namespaces and `...`) - maybe I could take a reverse truncation? I need to store a map of these so we don't reallocate any approach per-request, that's just being wasteful for something we can assume is fairly finite. I could just make name -> short name in a dictionary here I suppose. Let me play and see what looks good!

src/MiniProfiler.AspNetCore.Mvc/MvcDiagnosticListener.cs
NickCraver: When a request ends, we go through the tree and stop all running timers (it'd be the head), so no big deal. The timing may be inaccurate but...I can't imagine you'd care it's off really. If it's the best we can do, doubly fine with that - thoughts?

src/MiniProfiler.AspNetCore.Mvc/MvcDiagnosticListener.cs
NickCraver: Good call - added these in!

src/MiniProfiler.AspNetCore.Mvc/MvcDiagnosticListener.cs
rynowak: Seems fine, I just wanted to make sure you were aware of the possibility.

src/MiniProfiler.AspNetCore.Mvc/MvcDiagnosticListener.cs
NickCraver: Pushed up a new version here - may play with the caching a bit, just still catching up across the board after being out last week so a bit slower than usual. Thanks for all the comments here! Hopefully I'll be able to dogfood on Stack Overflow this week to get some testing in.

PR #483, opened 2020-05-15
Adding support to RavenDb 4.1 Storage on MiniProfiler

lillo42:
Adding support to RavenDb 4.1 Storage on MiniProfiler


NickCraver:
Heyyyy awesome, it looks like they're strongly naming the client now which means this is viable! That was the main blocker before :)

The compose and such is a great addition too - there's more bits I wanted to add there already so this is a great motivator. I'll try and dig in this weekend!

src/MiniProfiler.Providers.RavenDB/RavenDbStorage.cs
```diff
@@ -0,0 +1,335 @@
...
+            {
+                query = query.Where(x => x.Started >= start.Value);
+            }
+
+            if (finish.HasValue)
+            {
+                query = query.Where(x => x.Started <= finish.Value);
+            }
+
+            query.Take(maxResults);
```
JesperTreetop, line 98: Shouldn't this be `query = query.Take(maxResults);`?

PR #491, opened 2020-05-29
RavenDb - Max result was not respected

lillo42:


NickCraver:
This looks like it'd limit _before_ filtering, shouldn't the changes just be from `query.Take(..);` to `query = query.Take(...);`? The bug before was it just wasn't using the result :)

NickCraver:
Since this was fixed in #494, going to close this out :)

PR #494, opened 2020-06-04
Don't drop `Take` query instruction on the floor

JesperTreetop:
This looked like a bug to me, assuming that the query works similarly to a regular `IQueryable`/`IEnumerable`.


JesperTreetop:
For what it's worth, I just noticed this by looking at the pull request being mentioned on Twitter. [All bugs are shallow](https://en.wikipedia.org/wiki/Linus%27s_law), and all that.

PR #495, opened 2020-06-05
Use Razor Pages route value in MiniProfiler name

jeffpapp:
If you are using Razor Pages instead of controllers and views MiniProfiler shows all result names in the results list as `/`.  This is because `routeData.Values["controller"]` and `routeData.Values["action"]` are null.  

This change checks if `routeData.Values["pages"]` has a value and uses that for the MiniProfiler name if available.

I've also added a Razor Page to the Asp.Net Core 3 sample for testing

### DIFF
--- a/samples/Samples.AspNetCore3/Pages/RazorPagesSample.cshtml
+++ b/samples/Samples.AspNetCore3/Pages/RazorPagesSample.cshtml
@@ -1,0 +1,37 @@
+﻿@page
+
+@{
+    ViewData["Title"] = "Razor Pages Sample";
+}
+<div class="col-md-12">
+    <div class="page-header">
+        <h2>ASP.NET Core 3: Behold MiniProfiler in the top right!</h2>
+    </div>
+</div>
+<div class="row">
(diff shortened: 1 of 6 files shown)


NickCraver:
Ooh awesome - thank you! I'll pull to test this in a bit, I am hesitant about the check though, for example if someone has `"page"` in their routing at all (a common name), this would do the wrong thing - I think probably checking if controller has a value, then else if the page has a value, then else to unknown would be the best path here - thoughts on that logic path?

jeffpapp:
> Ooh awesome - thank you! I'll pull to test this in a bit, I am hesitant about the check though, for example if someone has `"page"` in their routing at all (a common name), this would do the wrong thing - I think probably checking if controller has a value, then else if the page has a value, then else to unknown would be the best path here - thoughts on that logic path?

That makes sense to me, maybe fallback to the url like the logic below this if the route values don't have `"controller"` or a `"page"`.  Do you want me to change it here or can you push a change?

NickCraver:
If cool with it - happy to push tonight after kiddos are in bed :) Thanks!

Issue #460, opened 2020-03-14
How to upgrade the Dapper.StrongName  version to 1.60.6


Nathan2020-0126:
There is a conflication with my existing project, how to solve it?
![image](https://user-images.githubusercontent.com/16734387/76692487-786eb100-6692-11ea-9275-945303ecc98b.png)

Nathan2020-0126:
`严重性	代码	说明	项目	文件	行	禁止显示状态
错误	CS0433	类型“DynamicParameters”同时存在于“Dapper.StrongName, Version=1.50.2.0, Culture=neutral, PublicKeyToken=e3e8412083d25dd3”和“Dapper, Version=1.60.0.0, Culture=neutral, PublicKeyToken=null”中	Mom.Web	D:\02_工作\00_Project\01_南方轴承\项目文件\备份\0315引入MiniProfiler之前 - 副本\Mom\Mom\Mom.Web\Initialization\SqlErrorStore.cs	187	活动
`

NickCraver:
I see, your difference is Dapper and Dapper.StrongName. There's no easy fix for this - it's quite painful.

The plan here is (for info: I also co-maintain Dapper) to strong name Dapper v3 and have an single, strongly named NuGet package. We aim to do that this year but with the world the way it is...honestly we don't now what the future holds for timelines.

PR #496, opened 2020-06-06
upgrade Dapper.StrongName to 2.0.35

ryuuc:


NickCraver:
Hi there! Can you please explain what problem you are trying to solve here? In general we rely in the lowest version of a library we need as to be the most compatible with the most consumers. MiniProfiler will upgrade to Dapper v3 at some point, but we do not require > 1.50.x in Dapper and want to be as compatible as possible.

Is there a specific issue you're hitting?

ryuuc:
lastest dapper referenced in my code,and then version conflict like this [https://github.com/MiniProfiler/dotnet/issues/460](url). I solved the problem by copy SqlServerStorage.cs to my project... you are right compatible is important,I will close this pr

PR #510, opened 2020-07-15
Update MySqlConnector to 1.0

bgrainger:
MySqlConnector 1.0 introduces a breaking change by changing the namespace: https://github.com/mysql-net/MySqlConnector/issues/824

This updates the dependency and fixes the code.


bgrainger:
The current version of MiniProfiler.Providers.MySql on NuGet ([v4.2.1](https://www.nuget.org/packages/MiniProfiler.Providers.MySql/4.2.1)) doesn't include this change.

Unfortunately MySqlConnector v1.0 was a binary-breaking change, so websites that are using MySqlConnector >= 1.0 can't also use MiniProfiler.Providers.MySql 4.2.1. Can an update be shipped (v4.3? v5.0?) so that there's the option to resolve the runtime conflict?

PR #516, opened 2020-08-01
Explicitly set MySQL connection string options

bgrainger:
Instead of relying on the defaults, ensure that MySqlConnector string options are set to the values that this library requires to (de)serialize data correctly:

* Force `DateTime` to be stored in UTC and returned as `DateTime` (not `MySqlDateTime`)
* Store `Guid` as `CHAR(36)`
* Disable user variables and connection reset (optimisation that's safe because no variables or temporary tables are used)
* Store `bool` as `TINYINT(1)`

This improves robustness if the user passes in a connection string with overridden defaults (e.g., if MySQL is being used by the application for other purposes and needs a different configuration).

### DIFF
--- a/src/MiniProfiler.Providers.MySql/MySqlStorage.cs
+++ b/src/MiniProfiler.Providers.MySql/MySqlStorage.cs
@@ -19,7 +19,7 @@ namespace StackExchange.Profiling.Storage
         /// Initializes a new instance of the <see cref="MySqlStorage"/> class with the specified connection string.
         /// </summary>
         /// <param name="connectionString">The connection string to use.</param>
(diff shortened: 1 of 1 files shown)


NickCraver:
Thanks @bgrainger! Given how often we call GetConnection() - thoughts on a `Dictionary<string, string>` for the translation so we don't create and iterate through a builder every call with the same results?

bgrainger:
Ouch--thanks for catching that. Yes, it only needs to be modified once; I'll just change it on the way to passing it to the base class.

bgrainger:
@NickCraver I force-pushed an update that only rewrites the connection string once.

NickCraver:
oh yeah even better - thanks! I'll get this in after tests pass :)

PR #521, opened 2020-08-24
Add ProfiledDbCommand.AlwaysWrapReaders

NickCraver:
This adds a path for always wrapping readers so that anyone inheriting from `ProfiledDbDataReader` and overriding functionality can depend on that by also overriding this new property to `true`.

### DIFF
--- a/src/MiniProfiler.Shared/Data/ProfiledDbCommand.cs
+++ b/src/MiniProfiler.Shared/Data/ProfiledDbCommand.cs
@@ -22,6 +22,12 @@ namespace StackExchange.Profiling.Data
         private bool _bindByName;
 
         /// <summary>
+        /// Whether to always wrap data readers, even if there isn't an active profiler on this connect.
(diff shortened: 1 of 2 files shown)


mgravell (approved):

jdaigle:
@NickCraver do you create subclasses of ProfiledDbConnection/DbCommand/DbDataReader? What sorts of things are you doing the subclasses?

The reason I ask is that I’m considering creating a subclass of ProfiledDbCommand to add some additional logging whenever a command is executed.

I’m thinking about something like what you described here: https://twitter.com/nick_craver/status/1011056299705020418?s=21. We’re having problems, again, that might be async or transaction related, and I want to capture more information in our exception logging related to the current state of the commands/connections.

mgravell:
@jdaigle we do indeed; it does a few things, including:

- better connection tracking than ADO-NET provides
- isolation level tweaks and optimisations

jdaigle:
@mgravell thanks! I’d love to learn more about what you’re doing, it’s very fascinating, but I understand if it’s asking too much.

PR #524, opened 2020-09-01
Fixes for casting exceptions when using SqliteStorage with local db

Nemina:
With SqliteStorage when using a local db file 

1. Added "IF NOT EXISTS" clause to create table script to prevent the exception when starting the project for the second time. 

2. Added decimal points to instances when missing to prevent Dapper mapping from throwing a cast exceptions

image

### DIFF
--- a/samples/Samples.AspNetCore2/Startup.cs
+++ b/samples/Samples.AspNetCore2/Startup.cs
@@ -122,7 +122,7 @@ namespace Samples.AspNetCore
                 dbContext.Database.EnsureCreated();
             }
             // For nesting test routes
-            new SqliteStorage(SqliteConnectionString).WithSchemaCreation();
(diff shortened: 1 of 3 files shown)


NickCraver:
Hey there! Thanks for the contribution here :) A few overall notes:
1. This creates the schema all the time (even if the user doesn't intend it) and removes their control of doing so - especially in a shared environment that's not desirable and why we have an extension method (it'd be a breaking change as well). On the double front, can you please illustrate an example where this fails? It's likely we should change how we're calling Dapper here, rather than the SQL. If you have a repro (or a failing test!) I'd be happy to take a look :)

NickCraver:
I'd like to fix whatever issue was hit here, but without more info I don't have anything to go off of - closing this out for now.

PR #525, opened 2020-09-09
Update AspDotNetCore.md to add Memory Cache

ceemafour:
Add required step to add services.AddMemoryCache() which is essential for MiniProfiler to work in .Net Core

### DIFF
--- a/docs/AspDotNetCore.md
+++ b/docs/AspDotNetCore.md
@@ -21,6 +21,9 @@ Install-Package MiniProfiler.AspNetCore.Mvc -IncludePrerelease
 public void ConfigureServices(IServiceCollection services)
 {
     // ...existing configuration...
+    
+    // Ensure Memory Cache is added for MiniProfiler to work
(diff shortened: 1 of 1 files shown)


NickCraver:
@ceemafour This should not be needed since 4.2.1 (see release notes: https://miniprofiler.com/dotnet/Releases) - are you still encountering a case where it is?

ceemafour:
Yes it’s still happening. Downloaded the nuget 4.2.1 today and ran into this issue.

NickCraver:
@ceemafour we should look at your config - what error is happening? Note from the examples - it should not be needed: https://github.com/MiniProfiler/dotnet/blob/main/samples/Samples.AspNetCore3/Startup.cs#L34 - they run just fine. I'd like too understand what's different about your scenario that it's not behaving, rather than adjusting the docs. There's a bug here if you still need this.

NickCraver:
Closing this out to cleanup!

PR #530, opened 2020-09-28
Step Extension

KSemenenko:
I started using the profiler in a WebAPI project and realized that it was very uncomfortable for me to write the name of a step with my hands.

So I propose an extension that will generate a step name.

``` cs
protected void SomeMethod()
{
    using (profiler.Step(profiler.GetDefaultStepName()))
    {
        // you can start profiling your code immediately
    }
}
//or
protected void SomeMethod()
{
    using (profiler.Step())
    {
        // you can start profiling your code immediately
    }
}
```

``` cs
string GetDefaultStepName(); 
string GetDefaultStepNameWithClassName();
string GetDefaultStepNameWithLine();
```
Only 3 methods which will generate type names:

``` cs
string GetDefaultStepName(); 
//SomeMethod
```
``` cs
string GetDefaultStepNameWithClassName();
// SomeController.SomeMethod
```

``` cs
string GetDefaultStepNameWithLine();
// SomeController.SomeMethod:59
```

and
``` cs
  using (profiler.Step())
// SomeMethod
```

### DIFF
--- a/samples/Samples.Mvc5.EFCore/Global.asax.cs
+++ b/samples/Samples.Mvc5.EFCore/Global.asax.cs
@@ -67,7 +67,7 @@ namespace Samples.Mvc5
                 profiler = MiniProfiler.StartNew();
             }
 
-            using (profiler.Step("Application_BeginRequest"))
+            using (profiler.Step(profiler.GetDefaultStepName()))
             {
                 // you can start profiling your code immediately
             }
--- a/samples/Samples.Mvc5.EFCore/Global.asax.cs
+++ b/samples/Samples.Mvc5.EFCore/Global.asax.cs
@@ -67,7 +67,7 @@ namespace Samples.Mvc5
                 profiler = MiniProfiler.StartNew();
             }
 
-            using (profiler.Step(profiler.GetDefaultStepName()))
+            using (profiler.Step())
             {
                 // you can start profiling your code immediately
             }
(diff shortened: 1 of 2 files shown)


NickCraver:
Hey there! Thanks for the contributions :) Ultimately, I don't think this is something that belongs in the base library, for a few reasons:
1. What's "default" or important varies and I've seen a lot of implementations.
2. This approach allocates every single call - that's `n` strings repeating the same work (though we could improve it a little with a cache based on a tuple key).
3. The nature of these is extensions, so that while default differs, you can create an extension that works _for you_ in your solution or library and not be bound by any conventions here.

For example, you can do this _outside_ the library:
```c#
/// <summary>
/// Does a Step with the location of caller as the label.
/// </summary>
public static IDisposable StepHere(
    this MiniProfiler profiler,
    [CallerMemberName] string memberName = "",
    [CallerFilePath] string sourceFilePath = "",
    [CallerLineNumber] int sourceLineNumber = 0) =>
    profiler?.Step($"{memberName} - {Path.GetFileName(sourceFilePath)}:{sourceLineNumber.ToString()}");
```

So I think these are great examples of what you _can_ do (with the allocation caveat), but I wouldn't want to package them in the main library since their intended implementation is already known to vary a lot - does that make sense? Thoughts?

KSemenenko:
hello!
thank you for your answer,
Of course I made myself extensions. 
And I thought it might be useful in the library :)
Then I will probably close this PR. thank you again for your wonderful library!

PR #537, opened 2020-10-01
Update MiniProfilerHandler.cs

prafullthorve:
Added If condition in Includes method to check path is not null.


NickCraver:
XMLdoc is appreciated! Is there a case where `path` _is_ `null` that you've hit?

NickCraver:
Going to close this out since the major change isn't reasoned here - if there _is_ a problem, let's please address that but the area to address such a thing is very likely upstream. But that is also assuming there's a problem we're solving...I can't tell from the changes and no follow-up here. If you see this and there's a reason, happy to re-open!

PR #541, opened 2020-12-02
Include Area in name for AspNetCore

Diego-T:
Currently on the AspNetCore version, only the Controller and Action are used in the profiler name.  This adds the Area too like in the Mvc5 version.


NickCraver (changes requested):
If we're adding area here, let's please do it in an `if/else` - note that concatenating a string is faster than a format which has been added here with interpolation. It's also faster to `.ToString()` the `area` object retrieved. So overall, we should have 2 lines of `profiler.Name`, one of each branch of the `if` checking if `.TryGetValue("area"...` succeeded.

Hope that makes sense - if not happy to push changes to this and absorb but wanted to give feedback on why the tweaks!

NickCraver (approved):
:+1: Thanks!

PR #544, opened 2021-02-07
MongoDB: Allow custom collection names for profiling results

### LINKED ISSUE #542: Allow MongoDB Storage custom collection name
Currently the MongoDB storage always stores in a collection called "profilers".
It should be possible to customize this by passing a custom name a second constructor parameter.

Code from 
https://github.com/MiniProfiler/dotnet/blob/main/src/MiniProfiler.Providers.MongoDB/MongoDbStorage.cs

Notice "profilers" hard coded in this code:

```
        public MongoDbStorage(string connectionString)
        {
            if (!BsonClassMap.IsClassMapRegistered(typeof(MiniProfiler)))
            {
                BsonClassMapFields();
            }

            var url = new MongoUrl(connectionString);
            var databaseName = url.DatabaseName ?? "MiniProfiler";

            _client = new MongoClient(url);
            _collection = _client
                .GetDatabase(databaseName)
                .GetCollection<MiniProfiler>("profilers");
        } 
```

Turnerj:
Fixes #542 

Part of me wanted to add a test to assert the optional collection name but with how the code is structured currently, it felt a little over the top.

### DIFF
--- a/src/MiniProfiler.Providers.MongoDB/MongoDbStorage.cs
+++ b/src/MiniProfiler.Providers.MongoDB/MongoDbStorage.cs
@@ -19,7 +19,8 @@ namespace StackExchange.Profiling
         /// Returns a new <see cref="MongoDbStorage"/>. MongoDb connection string will default to "mongodb://localhost"
         /// </summary>
         /// <param name="connectionString">The MongoDB connection string.</param>
-        public MongoDbStorage(string connectionString)
+        /// <param name="collectionName">The collection name to use in the database.</param>
+        public MongoDbStorage(string connectionString, string collectionName = "profilers")
         {
             if (!BsonClassMap.IsClassMapRegistered(typeof(MiniProfiler)))
             {
@@ -32,7 +33,7 @@ namespace StackExchange.Profiling
             _client = new MongoClient(url);
             _collection = _client
                 .GetDatabase(databaseName)
-                .GetCollection<MiniProfiler>("profilers");
+                .GetCollection<MiniProfiler>(collectionName);
         }
 
         private static void BsonClassMapFields()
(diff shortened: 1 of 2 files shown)


NickCraver (changes requested):
Looks like the build is broken (we need to update the version of Nerdbank to latest), can take a look at that when on a laptop tonight hopefully. On the PR though: this is a breaking change, since optional parameters are on the caller in .NET at compile time. This will break users depending on binary compatibility (e.g. any lib calling this method in our lib).

This is okay on a major version break to do, but not friendly and against SemVer on any minor release - to resolve this, we need to have an overload so that the old signature calls this method, rather than an optional param. So overall: 2 methods, one with a non-optional param, and the old calling it with the new name. In terms of code diff, I recommend making the *old* constructor the new method in the diff if that makes sense (e.g. take this without the optional default, then go add the old method back for compatibility above it).

On the test front: we can ensure this works without major changes. Take a peek at the SqlServer and related tests, see how they're passing in things like `"MPTest" + TestId`? We should do that on Mongo as well, just so that we ensure there's no dependency on the default name somewhere and this indeed keeps working.

Turnerj:
Thanks for the feedback! While I've done a lot of .NET programming, I haven't yet come across things like binary compatibility problems (or maybe I have and just blamed something else) so knowing about it more clearly like in your description is super helpful.

I've implemented the constructor the way I believe you were referring to - keeping the old one but pushing its values through the new one. I've also added the test like you described, following the same pattern as the SQL one.

If it is helpful for you, I created a PR (#545) for updating Nerdbank.GitVersioning and like you thought, that fixes the issue with the build. If you want to merge that in, I can rebase on top of that or something.

Let me know if there are any other issues etc and I'll gladly get them sorted!

NickCraver:
I think this will be good after build checks - can you please merge in `main` in now that NerdBank is fixed up? (thanks for that!)

Turnerj:
No problem! Rebased onto `main` and everything is green!

Somewhat as an aside, is there anything I can do with this repo to help you maintain it? I'm thinking of opening a few PRs (some for adding/fixing things people have raised as issues, some for other changes/fixes I've noticed) but I don't want to overwhelm you with heaps more things to look at. Or if you had ideas for a next major version, I'd be happy to tinker around with something like that.

NickCraver (approved):
Also good - thank you!

PR #549, opened 2021-02-25
InlineFormatter: Optionally support spacing around commas

### LINKED ISSUE #539: Spaces inserted after commas in command text
Found that (even without a SqlFormatter configured) the SQL as it appears in MiniProfiler differs from the original command text; extra spaces are found after commas. The spaces are inserted [here](https://github.com/MiniProfiler/dotnet/blob/main/src/MiniProfiler.Shared/Internal/IDbCommandExtensions.cs#L41).

This is a problem for our team as we use MiniProfiler's output to diagnose poorly performing queries, and need to ensure same query plan is used (and thus exact same SQL) when we execute in SSMS.

As a workaround we're running the same regex on the SQL before it is executed.

Turnerj:
Fixes #539 

Default is enabled, like the existing GetReadableCommand logic.

### DIFF
--- a/docs/Releases.md
+++ b/docs/Releases.md
@@ -5,6 +5,10 @@ layout: "default"
 ### Release Notes
 This page tracks major changes included in any update starting with version 4.0.0.3
 
+#### Unreleased
+- **New**:
+  - Added an option to control `SpacesAfterCommas` to `InlineSqlFormatter` and `SqlServerFormatter` ([#549](https://github.com/MiniProfiler/dotnet/pull/549) - thanks [Turnerj](https://github.com/Turnerj))
+
 #### Version 4.2.1
 - **New**:
   - Added RavenDB Storage provider ([#483](https://github.com/MiniProfiler/dotnet/pull/483) - thanks [@lillo42](https://github.com/lillo42)!)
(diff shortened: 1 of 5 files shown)


Turnerj:
I don't like that I have 2 new declarations for that same Regex for adding the space around the commas - probably better having that sit on another class as a static or something that all 3 areas can now share. Also I kinda want to make the `InlineFormatter` and the `SqlServerFormatter` be structured more similarly - not just that the `InlineFormatter` uses a private static vs the `SqlServerFormatter` using a public property, probably can tidy up the `InlineFormatter` to be more structured like the `SqlServerFormatter`.

Still need to write the tests though.

NickCraver:
Sorry I missed this one - FWIW I'd rename here to just "SpaceAfterComma" so it's very clear what it does. With any formatter, people will want to turn on and off specific things. But also, yeah...the layout isn't ideal is it. I see the pickle because of the existing interface and the public method though. When next at a PC, lemme play around with this one and see if we can land things in another location - definitely agree with the intent, just seeing how we can make this more maintainable and shared.

Turnerj:
Yep, no problem - no need to apologise! I'll do a quick rename of the property to "SpaceAfterComma" with related tests and inline docs being updated just so that is fixed. You should be able to push commits to my branch if you've worked out how you want to restructure it.

If we can work something out for v4, even if its not ideal structure-wise, we could look at a bigger redesign for the next major version to fix it up more nicely.

NickCraver (approved):
:+1: Been heads down on StackExchange.Redis lately - getting back to other PRs, thanks for this!

Turnerj:
No problem! I wasn't sure if you were still wanting to look into your previous comment about restructuring where the Regex is. 🙂

Issue #350, opened 2018-12-27
MiniProfiler no showing up in mvc5


dashenxian:
A few days ago I added Miniprofiler to the MVC5 project and worked well, but after a few days I found that it was not working, the page no longer showed Miniprofiler, I tried to create a new MVC5 project, and according to https:// Miniprofiler.com/dotnet/aspdotnet's guidance method adds Miniprofiler and sets the Global.asax, and adds "Miniprofiler.current?.Renderincludes ();" to the _layout.cshtml but without any effect

NickCraver:
Can you show your Global setup and Razor code here?

dashenxian:
`   
protected void Application_Start()
{
            ConfigMiniProfiler();
//....other code
}
 private static void ConfigMiniProfiler()
        {
            MiniProfiler.Configure(new MiniProfilerOptions
            {
                // Sets up the route to use for MiniProfiler resources:
                // Here, ~/profiler is used for things like /profiler/mini-profiler-includes.js
                RouteBasePath = "~/profiler",

                // Example of using SQLite storage instead
                //Storage = new SqliteMiniProfilerStorage(ConnectionString),

                // Different RDBMS have different ways of declaring sql parameters - SQLite can understand inline sql parameters just fine.
                // By default, sql parameters will be displayed.
                //SqlFormatter = new StackExchange.Profiling.SqlFormatters.InlineFormatter(),

                // These settings are optional and all have defaults, any matching setting specified in .RenderIncludes() will
                // override the application-wide defaults specified here, for example if you had both:
                //    PopupRenderPosition = RenderPosition.Right;
                //    and in the page:
                //    @MiniProfiler.Current.RenderIncludes(position: RenderPosition.Left)
                // ...then the position would be on the left on that page, and on the right (the application default) for anywhere that doesn't
                // specified position in the .RenderIncludes() call.
                PopupRenderPosition = RenderPosition.Right,  // defaults to left
                PopupMaxTracesToShow = 10,                   // defaults to 15

                // ResultsAuthorize (optional - open to all by default):
                // because profiler results can contain sensitive data (e.g. sql queries with parameter values displayed), we
                // can define a function that will authorize clients to see the JSON or full page results.
                // we use it on http://stackoverflow.com to check that the request cookies belong to a valid developer.
                ResultsAuthorize = request => request.IsLocal,

                // ResultsListAuthorize (optional - open to all by default)
                // the list of all sessions in the store is restricted by default, you must return true to allow it
[...]

PR #555, opened 2021-04-15
Instructions on actually rendering the mini profiler

Bouke:
I found the installation instructions confusing, and [this issue](https://github.com/MiniProfiler/dotnet/issues/350) provided some insights. I think it would be valuable to make these instructions part of the documentation.

### DIFF
--- a/docs/AspDotNet.md
+++ b/docs/AspDotNet.md
@@ -104,7 +104,7 @@ protected void Application_EndRequest()
   </html>
   ```
 
-* Edit your `Web.config` to serve the resources, the `path` attribute should match `RouteBasePath`:
+* Depending on existing config, you may need to edit your `Web.config` to serve the resources, the `path` attribute should match `RouteBasePath`:
   ```xml
   <configuration>
     <system.webServer>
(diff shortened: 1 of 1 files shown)


docs/AspDotNet.md
```diff
@@ -95,8 +95,28 @@ protected void Application_EndRequest()
...
+* Edit your `Views\Shared\_Layout.cshtml` to render the MiniProfiler:
+  ```
+  @using StackExchange.Profiling
+  ...
+  @(MiniProfiler.Current?.RenderIncludes())
+  </body>
+  </html>
+  ```
+
+* Edit your `Web.config` to serve the resources, the `path` attribute should match `RouteBasePath`:
```
NickCraver, line 107:
```suggestion
* Depending on existing config, you may need to edit your `Web.config` to serve the resources, the `path` attribute should match `RouteBasePath`:
```

NickCraver (approved):
This is looking good - thank you! The Web.config bit may not be needed, so I suggest we tweak that to reduce confusion then good to go :)

Issue #578, opened 2021-10-13
Sqlite Provider InvalidCastException


MarkZither:
When using the Sqlite Provider i encountered the following error.

```
An unhandled exception occurred while processing the request.
InvalidCastException: Unable to cast object of type 'System.Int64' to type 'System.Double'.
Deserialize0693dad1-08a7-4b56-b158-09ed5699f3a7(IDataReader )

DataException: Error parsing column 5 (StartMilliseconds=120 - Int64)
Dapper.SqlMapper.ThrowDataException(Exception ex, int index, IDataReader reader, object value) in SqlMapper.cs, line 3665
```

This is similar to the final message in the PR #221 

It seems to stem from the dynamic [datatypes of Sqlite](https://www.sqlite.org/datatype3.html).

MarkZither:
Is my change ok to be accepted?

meum:
I also ran into this one today, would be great to get the PR merged

NickCraver:
I happen to be working on this after a long break just this week - this will be in a 4.3 release soon :)

NickCraver:
Going to close this our since #582 is merged in - look out for a 4.3 release on NuGet soon, but already on MyGet now :)

PR #582, opened 2021-11-07
Fix/sqlite datatypes

MarkZither:
Fix for #578 
Changed all DECIMAL columns to REAL to avoid an issue with dynamic datatypes in sqlite as discussed in https://stackoverflow.com/questions/39980840/sqlite-using-dapper-error-parsing-column-unable-to-cast-object-of-type-system


DrLeh:
I cloned the source code of this storage to my project with this change in it and it worked fine.

NickCraver (approved):
Sorry I didn't get to this until now - getting in for the 4.3 release, thank you!

PR #586, opened 2021-12-05
Add color-scheme light/dark to the  themes

xt0rted:
This adds [`color-scheme: light/dark`](https://drafts.csswg.org/css-color-adjust/#color-scheme-prop) to the themes so the browser uses the correct version/color of the scroll bar on these elements.

Config| Before | After
-- | -- | --
Light on Dark | image | image
Dark on Light | image | image
Dark | image | image


NickCraver (approved):
TIL this exists, thank you!

PR #617, opened 2022-08-08
Refactoring MiniProfilerBaseOptionsExtensions.ExpireAndGetUnviewedAsync

profet23:
Refactoring `MiniProfilerBaseOptionsExtensions.ExpireAndGetUnviewedAsync` to utilize a new overload of `IAsyncStorage.SetViewedAsync` that accepts multiple ids.
Adding `PostgreSqlStorage` implementation that can set multiple profilers as viewed in a single query.

This query is called a lot, so trying to reduce database round trips.

### DIFF
--- a/src/MiniProfiler.AspNetCore/Storage/MemoryCacheStorage.cs
+++ b/src/MiniProfiler.AspNetCore/Storage/MemoryCacheStorage.cs
@@ -249,5 +249,18 @@ namespace StackExchange.Profiling.Storage
             SetViewed(user, id);
             return Task.CompletedTask;
         }
+        
+        /// <summary>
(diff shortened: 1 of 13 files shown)


NickCraver:
For this to happen, it'll need to wait for a 5.x release due to breaking a primary interface here for current implementers. I'll take a peek and surface looks sound, but it is breaking so want to give an early heads up on timing expectations.

profet23:
> For this to happen, it'll need to wait for a 5.x release due to breaking a primary interface here for current implementers. I'll take a peek and surface looks sound, but it is breaking so want to give an early heads up on timing expectations.

It should be backwards compatible. I don't believe there are any breaking changes.

PR #618, opened 2022-08-09
New IAdvancedAsyncStorage interface

profet23:
Taking the comments from: https://github.com/MiniProfiler/dotnet/pull/617

In order to not break existing implementations of `IAsyncStorage`, this PR proposes the new interface `IAdvancedAsyncStorage` which inherits IAsyncStorage and seeks to provide some additional functionality.

The idea would be to merge IAsyncStorage and IAdvancedAsyncStorage in a future release.

### DIFF
--- a/src/MiniProfiler.Providers.PostgreSql/PostgreSqlStorage.cs
+++ b/src/MiniProfiler.Providers.PostgreSql/PostgreSqlStorage.cs
@@ -12,7 +12,7 @@ namespace StackExchange.Profiling.Storage
     /// <summary>
     /// Understands how to store a <see cref="MiniProfiler"/> to a PostgreSQL Server database.
     /// </summary>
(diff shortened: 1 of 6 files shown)


profet23:
I'd like to make the new IAdvancedAsyncStorage internal here, but I'm honestly not sure how to implement InternalVisibleToAttribute with the dll signatures and build process. 

Input welcome.

NickCraver:
Hey @profet23 sorry I should have been clearer on other PR: it's not an issue, just that it requires a major release. I have no issues rounding up current items and doing a 5.x release (needs testing with .NET 6, etc.). If we split the interface, then downstream implementors may not ever know about another interface and adding new ones indefinitely I don't think is a sound strategy :)

I haven't had the time allocation lately to look at these but based on surface: let's stick with your original and version appropriately.

PR #637, opened 2023-02-09
Overall library update: Drop netcoreapp3.1/net5.0, move to net6.0

NickCraver:
The versions of AspNetCore that previously supported netstandard2.0, net461, and even netcoreapp3.1 are all unsupported now, so let's move to the LTS. This will light up builds again and we can get a new version out.

Bumping version to 4.3 since it's a dependency drop just to better indicate.


NickCraver:
@mgravell If you could sanity check this I'd appreciate it - the file list looks big, but it's a massive delete in there - actual delta is a small subset of that. I think this drop makes sense given the support lifecycle, but would love your thoughts!

NickCraver:
@DamianEdwards Sanity checking: does this TFM set (for maximum compat.) look correct to you, or different advice? I'm not daily using ASP.NET 6+ libs unfortunately, so wanted to check before doing a release here.

DamianEdwards:
From an ASP.NET Core POV, where versions 2.1 on supported NetFx versions, and versions 6.0 and 7.0 on .NET 6.0 and .NET 7.0 respectively, are supported, this looks fine. I noted the test project still seems to target net461 though. Is there a specific reason not to align that with the min supported NetFx version of 471 now?

NickCraver:
@DamianEdwards Thanks for eyes! Nope, no strong reason other than aligning to what the library supports so we don't have a random TFM only on the tests. I think in reality nothing matters there it's all going to be running on a .NET 4.8 box anywhere on consequence and API usage is minimal.

PR #640, opened 2023-02-10
Nullable Reference Types: Let's goooooooooo

NickCraver:
Adding more checking for everything in the form of NRTs. This may warrant a version bump, I need to go back and look at the API surface area and add those analyzers in to decide.


m0sa:
:shipit:

PR #650, opened 2023-05-23
Add support for hooking in additional instrumentation around Timings

m0sa:
Most of the code of _a project that I'm working on_ is instrumented via MiniProfiler's extension methods. In order to get the MiniProfiler instrumented sections/traces into 3rd party APM tools, we'd either need to hijack all the MiniProfiler extension method combinations, or add yet another `using` around everything we instrument via MiniProfiler. The latter is particularly impractical for the `.Inline(() => ....)` extension methods etc.

This is just a little something to get a discussion going. How opposed would you be to something like this?


m0sa:
@NickCraver we don't use ActivitySource / Listener by default either, and doing having to use a Listener just to forward it to something else is indeed an overkill. I think this more minimal approach would be more flexible for folks to hook in whatever they need, while also being more lightweight.

NickCraver:
@m0sa I dig it, because later this could be a one-liner to set to a static Func that could hook Activity the same way and we could bundle that in-box too if no additional dependencies. Let's get this in :)

PR #658, opened 2023-09-04
Attempt Fix #505 SQL Timings Not Displaying - for ProfiledDbDataAdapt…

### LINKED ISSUE #505: SQL Timings Not Displaying
I am implementing MiniProfiler based on the Samples.Mvc5 project included in the repo and am having an issue with the display of SQL timings. I am curious if something might be off in my setup, but I am not sure exactly what that might be.

Here is an example of loading the homepage, and I am confused why the SQL timings and percentage all show as 0.0:
image

However, if I actually click on the sql timings I get this view, which does seem to indicate that each SQL call does have timings associated with it:
image

Can you help direct me as to why I might not be seeing them aggregated in the initial view, or where I might start looking to gather more info?

urbanekv:
…er.Fill() called with a dataset

Hi @NickCraver , after some exploring, I dared to post this draft. It helped us to solve the problem with missing stats - issue #505 

In our project, we use ProfiledDbDataAdapter with the DataSet object passed to the FIll() method . It helped us to get rid of not-well-closed timings. The totals are now filled correctly.

### DIFF
--- a/src/MiniProfiler.Shared/Data/ProfiledDbDataAdapter.cs
+++ b/src/MiniProfiler.Shared/Data/ProfiledDbDataAdapter.cs
@@ -91,7 +91,7 @@ namespace StackExchange.Profiling.Data
             }
             finally
             {
-                _profiler.ExecuteFinish(cmd, SqlExecuteType.Reader, TokenReader);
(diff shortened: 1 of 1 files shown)


tvavrda:
@NickCraver can we ask for a status of this PR? Is there anything we can do? Thanks.

PR #665, opened 2023-10-13
Fix problem with showing async method profile

curllog:
In this example we don't have a result for `TestAsync() ` 

since method is async the result of **MiniProfiler.Current is null** in main method so we have to print result inside TestAsync Method


NickCraver (approved):
Thanks for this!

PR #666, opened 2023-11-14
Fix method name of async ResultsListAuthorizeAsync method in documentation.

afuersch:
Fix method name of async `ResultsListAuthorizeAsync` method in documentation.


NickCraver:
Thanks for this!

Issue #659, opened 2023-09-05
MiniPorfiler Keep repeating the requests


Abanoub98:
I'm using a mini profiler to examine my project performance however, I have a problem where the requests in profiler/results-index is repeated as I am using an infinite  loop 

My solution consists of  Asp.net core web API for the back end and Blazor for the view 

here is the profiler code in program.cs inside the Api project

```
builder.Services.AddMiniProfiler(
  options =>
  {
      options.RouteBasePath = "/profiler";
      (options.Storage as MemoryCacheStorage)!.CacheDuration = TimeSpan.FromMinutes(60);
      options.SqlFormatter = new StackExchange.Profiling.SqlFormatters.InlineFormatter();
      options.ColorScheme = StackExchange.Profiling.ColorScheme.Auto;
  }).AddEntityFramework();
```

![Screenshot 2023-09-05 043434](https://github.com/MiniProfiler/dotnet/assets/54473504/ac036e2a-38d0-46a6-8a69-c8c01284dfeb)

NickCraver:
I'm not sure I'm clear on the report/ask - it looks like you indeed have those requests all distinct with times, etc. MiniProfiler is only reporting what it sees here.

Abanoub98:
@NickCraver 
No, it keeps repeating the requests while they come only one time in the network tab and I debugged it many times and tested it in the logger.
All the requests is repeated every 2 seconds

NickCraver:
@Abanoub98 I agree that seems to be happening, but what does that have to do with MiniProfiler? It's only recording the requests it sees.

Abanoub98:
@NickCraver are there any alternatives to it? or any other options I can add? maybe the order of lines in the program.cs?

NickCraver:
@Abanoub98 Again, I think you're chasing the wrong thing here. You're asking  "why are there duplicate requests?", which is absolutely the question you should be asking - but you seem to be focused on MiniProfiler _causing_ them, rather than what it's almost certainly doing: _reporting they exist_. I cannot say why you're fetching every 2 seconds, my assumption would be errant JavaScript somewhere along the way causing reloads.

VitaliySmolyakov:
That must be same problem as https://github.com/MiniProfiler/dotnet/issues/653. I faced it too :(

klyakh:
I'm experiencing the same problem.

[...]

PR #550, opened 2021-03-24
Update MiniProfilerMiddleware.cs

spaduret:
fix trace records ordering

PR #668, opened 2024-02-10
Fix #653, #659 - Fix UI to work with most recent first

### LINKED ISSUE #653: polling use first instead of last
It appears that there is an issue with the JavaScript component responsible for polling the next result on the results-index page. The page is arranged in descending order by time, with the most recent item at the top. However, the last ID used for polling is the least recent one, causing the page to continuously grow larger as it retrieves n-1 results with each polling iteration.

image

Most likely the error is here https://github.com/MiniProfiler/dotnet/blob/a009d55a19050c8b88a3ea0c357d54ad192763b5/src/MiniProfiler.Shared/ui/lib/MiniProfiler.ts#L439C57-L439C64

it should be oldData[0]

Note even with that fix we will end up with the wrong result displayed as it will display all the new results at the bottom...

that's what we currently would see with 3 items:

```
first load
3
2
1
polling last id=1
3
2
polling last id = 2
3
polling last id = 3
polling last id = 3
polling last id = 3
```
if we use oldData[0]
this would work like that:
```
first load
3
2
1
polling last id = 3
polling last id = 3
polling last id = 3
```
but as soon as we get a new id 4 the polling will push it at the end and the page table will now display
```
3
2
1
polling last id = 3
4
```

RobertSmits:
Fixes both #653 and #659 by updating the UI to work with the recent-trace-first sorting introduced in #550

### DIFF
--- a/src/MiniProfiler.Shared/ui/lib/MiniProfiler.ts
+++ b/src/MiniProfiler.Shared/ui/lib/MiniProfiler.ts
@@ -430,13 +430,13 @@ namespace StackExchange.Profiling {
   <td colspan="3" class="mp-results-none">(no client timings)</td>`) + `
 </tr>`);
                         });
-                        document.querySelector('.mp-results-index').insertAdjacentHTML('beforeend', html);
+                        document.querySelector('.mp-results-index').insertAdjacentHTML('afterbegin', html);
                         const oldId = id;
                         const oldData = data;
                         setTimeout(() => {
                             let newId = oldId;
                             if (oldData.length > 0) {
-                                newId = oldData[oldData.length - 1].Id;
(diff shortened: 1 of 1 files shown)


axylophon:
@NickCraver : could you please look at this PR as the UI is currently broken...

afuersch:
@NickCraver Is there any chance to merge this PR soon? The UI is currently broken ...

LEIRONGHUA:
Does this PR have the latest progress?  The UI is Currently Broken, can no longer be used normally

NickCraver (approved):

NickCraver:
Thanks for the fix here! This will be int he 4.5.x update.

PR #669, opened 2024-04-09
Allow additional headers

Breederveld:
Allow the client code to provide additional headers for custom implementations.
Specifically the goal for us is to be able to include authentication using a bearer token, but this solution should cover other custom cases as well.


NickCraver (approved):
I see no harm in this, if it helps then 👍, thanks!

PR #681, opened 2025-01-06
fix: MiniProfiler.Current.RenderIncludes generates invalid html

rikbosch:
The `RenderIncludes`  method renders invalid html.

The internal `Render` method adds an extra `"` after the `data-scheme`  attribute.

image


sklivvz (approved):

NickCraver (approved):
Thanks for the fix!

PR #686, opened 2025-03-25
Bug Fix: SQL Profiling Network Time

### LINKED ISSUE #671: SQL Profiling: Network Time not assigned to SQL cost/time
The only documentation I can see around profiling SQL is here: https://miniprofiler.com/dotnet/HowTo/ProfileEFCore, and here https://miniprofiler.com/dotnet/HowTo/ProfileSql. It's not clear if these two approaches are equivalent, but the following issue uses the former. 

The documentation does not specify what to expect if you have a "fast" SQL query which spends a lot of time transferring data accross the network. 

For example let's say we run: 

```tsql
SELECT *
FROM [Table]
```

Within a code block like:

```c#
using (Profiler.Step("Materialise Data"))
{
    unit.GetTable<Table>().Tolist();
}
```

This is a simple query, which SQL quickly runs, and Miniprofiler sat on an instance right next to the DB would show a total time close to the query time. 

Now let's say we're hundreds of miles away from the DB. The time it takes to run the query _on SQL server_ is still fast, but the overall query time including network transit is slow. 

Where would you expect MiniProfiler to allocate that time? As far as I can tell, it does not appear in the SQL time, but makes it appear like the object materialisation is taking up the time: 

image

If I run the exact same SQL query from SQL Server Management Studio, it takes several seconds (not 20ms). 

Ideally we could separate out SQL query time from SQL network time, but failing that should the documentation be clearer on what time is being calculated and returned here?

dazbradbury:
This fixes: https://github.com/MiniProfiler/dotnet/issues/671

In our configuration `data.Result` is never a `RelationalDataReader`, but is always of type `SqlDataReader` (`System.Data.Common.DbDataReader`). This fixes the conditional so the profiling is logged correctly for both types of data reader.

Before: 

image

After:

image

### DIFF
--- a/src/MiniProfiler.EntityFrameworkCore/RelationalDiagnosticListener.cs
+++ b/src/MiniProfiler.EntityFrameworkCore/RelationalDiagnosticListener.cs
@@ -63,7 +63,7 @@ namespace StackExchange.Profiling.Data
                 if (val is CommandExecutedEventData data && _commands.TryRemove(data.CommandId, out var current))
                 {
                     // A completion for a DataReader only means we *started* getting data back, not finished.
-                    if (data.Result is RelationalDataReader)
+                    if (data.Result is RelationalDataReader or SqlDataReader)
                     {
                         _readers[data.CommandId] = current;
                         current.FirstFetchCompleted();


NickCraver (approved):
Thanks for this!

dazbradbury:
Thanks for merging.

Do you know roughly when there will be a new release that includes this fix? Just so I can tee up removing our custom code.

dazbradbury:
@NickCraver - Sorry to chase on this, any update on when the nuget package will be updated?

dazbradbury:
@NickCraver - Sorry to chase again on this, any update on when the nuget package will be updated so we can remove our custom handler and use the default package?

jdaigle:
Hey @NickCraver, any chance you could publish updated package versions with this fix?

